PR #2, opened 2020-12-07
chore: add basic draco python files

domoritz:
Just copying files https://github.com/uwdata/draco and adding python versions of the vega/compassql converters. Still needs the rest of the Python setup.


harshil93r (approved):
LGTM

PR #3, opened 2020-12-08
Setup

harshil93r:


domoritz:
I'm converting this to a draft. Let me know when it's ready for a review.

draco/cli.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python3
```
domoritz, line 1: I removed the CLI on purpose since I think we should use the Python interfaces instead.

requirements.txt
```diff
@@ -0,0 +1,9 @@
+ansunit
```
domoritz, line 1: We should probably remove this if we use poetry

setup.py
```diff
@@ -0,0 +1,36 @@
...
+    name="draco",
+    version=__version__,
+    description="Visualization recommendation using constraints",
+    long_description=long_description,
+    author="Dominik Moritz, Chenglong Wang",
+    author_email="domoritz@cs.washington.edu, clwang@cs.washington.edu",
+    license="BSD-3",
+    url="https://github.com/cmudig/draco2",
+    packages=["draco"],
+    entry_points={"console_scripts": ["draco=draco.cli:main"]},
```
domoritz, line 23: When we remove the CLI, this should go

requirements.txt
harshil93r: I'm having issues with poetry so used `pip install -r requirements` only.

But this ansunit is for unit test in answer programming and poetry is for environment management, so they're independent I think. Let me know if otherwise.

.github/workflows/test.yaml
```diff
@@ -1,36 +1,40 @@
...
+          which conda
+          conda --version
+
+
       - shell: bash -l {0}
         run: |
           pip install mypy
           pip install wheel
 
-      - name: Python Poetry Action
```
domoritz: No more poetry? How do we install dependencies and create a package?

README.md
```diff
@@ -1,5 +1,7 @@
 # Draco 2
 
+![Test](https://github.com/harshil93r/draco2/workflows/Test/badge.svg)
```
domoritz, line 3: Should point to cmudig

draco/cli.py
domoritz: The goal here is to support more diverse use cases. CLIs are usually limited to specific applications and don't allow for much flexibility. However, we want Draco to be used for the recommendation of charts given various inputs, enumeration of a space of possible charts, checking specs, etc.

requirements.txt
domoritz: One issue with that is that we don't fix versions of dependencies. We will eventually need a better setup but I am okay with merging a pull request and we can iterate later.

domoritz (changes requested):

.github/workflows/test.yaml
harshil93r: I was thinking basic `pip install -r ` and to build the package `python setup.py build`

requirements.txt
```diff
@@ -0,0 +1,9 @@
+ansunit==0.1.7
```
domoritz, line 1: No, I don't want to fix the direct dependencies. I want to lock all dependencies but be flexible with the direct dependencies. There should be a separate lockfile.

domoritz (changes requested):

draco/cli.py
harshil93r: sure, will be removing this file then.

domoritz:
done in master mostly

PR #76, opened 2021-04-22
refactor: add pyright type checker and update code

### LINKED ISSUE #68: Consider pyright for ci
https://github.com/microsoft/pyright

domoritz:
fixes #68


ShreyaBM (approved):
If we are switching to pyright, will I have to modify my code such that every variable has a specified type?

domoritz:
> If we are switching to pyright, will I have to modify my code such that every variable has a specified type?

No, not much should change.

PR #100, opened 2021-06-11
feat: support getting the top k models from clingo

### LINKED ISSUE #58: Add method to get top-k models from clingo
See https://github.com/uwdata/draco/blob/master/asp/topk-py.lp. Should be added to https://github.com/cmudig/draco2/blob/main/draco/run.py

domoritz:
closes #58


ShreyaBM (approved):
`run_clingo()` now has the functionality to return the requested number of models for an optimization problem in decreasing order of optimality.
Number of models returned = min(total models, requested models)

PR #183, opened 2022-01-15
feat: port some hard constraints from Draco 1

JunranY:


draco/asp/hard.lp
```diff
@@ -2,10 +2,48 @@
...
+    not attribute((mark,type),M,text),
     attribute(has_channel,M,text).
+
+% @hard(bin_and_aggregate) Cannot bin and aggregate.
+violation(bin_and_aggregate) :-
+    attribute((encoding,binning),E,_),
+    attribute((encoding,aggregate),E,_).
+
+% @hard(no_encodings) There has to be at least one encoding.
+violation(no_encodings) :- not entity(encoding,_,_).
```
domoritz, line 19: I think we can improve this and say there has to be at least one encoding for every mark.

draco/asp/hard.lp
```diff
@@ -2,10 +2,48 @@
...
+
+% @hard(repeat_channel) Cannot use single channels twice.
+violation(repeat_channel) :-
```
domoritz, line 22: I think this is wrong now since we can actually reuse the same encoding in differnet marks or different views.

draco/asp/hard.lp
```diff
@@ -2,10 +2,48 @@
...
+    attribute((encoding,channel),E1,C),
+    attribute((encoding,channel),E2,C),
+    E1 != E2.
+
+% @hard(row_no_y) Don't use row without y. Just using y is simpler.
+violation(row_no_y) :-
```
domoritz, line 28: This is also incorrect with multiuple views. Let's maybe not work on row encodings for now since I'm sot sure how exactly we will encode it.

draco/asp/helpers.lp
```diff
@@ -3,4 +3,4 @@
 % @helper(has_channel) Whether a mark has an encoding for a channel.
 attribute(has_channel,M,C) :-
     entity(encoding,M,E),
-    attribute(channel,E,C).
```
domoritz, line 6: Nice catch!

domoritz (changes requested):

draco/asp/hard.lp
JunranY: modified

draco/asp/hard.lp
JunranY: modified

draco/tests/test_hard.py
```diff
@@ -86,3 +86,243 @@ def test_text_channel_without_text_mark():
...
+    entity(encoding,root,e1).
+    attribute((encoding,channel),e1,x).
+    attribute((encoding,binning),e1,10).
+    attribute((encoding,aggregate),e1,count).
+
+    :- violation(_).
+    """
+    )
+
+    # assert not is_satisfiable(
```
domoritz, line 122: remove commented out code or fix it

draco/asp/hard.lp
```diff
@@ -9,3 +9,41 @@ violation(text_mark_without_text_channel) :-
...
+    entity(encoding,M,E2),
+    attribute((encoding,channel),E1,C),
+    attribute((encoding,channel),E2,C),
+    E1 != E2.
+
+% @hard(encoding_no_field_and_not_count) All encodings (if they have a channel) require field except if we have a count aggregate.
+violation(encoding_no_field_and_not_count) :-
+    not attribute((encoding,field),E,_),
+    not attribute((encoding,aggregate),E,count),
+    entity(encoding,_,E).
```
domoritz, line 36:
```suggestion
violation(encoding_no_field_and_not_count) :-
    entity(encoding,_,E),
    not attribute((encoding,field),E,_),
    not attribute((encoding,aggregate),E,count).
```

is slightly more readable to me

draco/tests/test_hard.py
```diff
@@ -86,3 +86,243 @@ def test_text_channel_without_text_mark():
...
+
+    assert not is_satisfiable(
+        p
+        + """
+    entity(encoding,root,e1).
+    attribute((encoding,channel),e1,x).
+    attribute((encoding,binning),e1,10).
+    attribute((encoding,aggregate),e1,count).
+
+    :- violation(_).
```
domoritz, line 118: I have an idea here. If we remove this line and check for violations instead of satisfiability, and test whether the list of violations is what we expect, we should have much more robust and precise tests. 

What do you think? Do you want to send a follow up pull request to make this improvement? 

```suggestion
```

domoritz (approved):
Looks great. Just one small fix.

draco/tests/test_hard.py
JunranY: Sure. I'm going to send a new pull request for this.

PR #188, opened 2022-01-20
feat:change test_hard format

JunranY:
referring to this suggestion: https://github.com/cmudig/draco2/pull/183#discussion_r788896536


draco/tests/test_hard.py
```diff
@@ -1,90 +1,324 @@
...
+
+    assert (
+        get_violations(
+            """
+    attribute((encoding,channel),e0,x).
+    attribute((encoding,channel),e1,y).
+
+    """
+        )
+        == []
```
domoritz, line 156: this can still be `is_satisfiable` as that's equivalent to checking for no violations but faster

draco/tests/test_hard.py
```diff
@@ -1,316 +1,313 @@
...
-        + """
-    entity(encoding,root,1).
+    assert (
+        get_violations(
+            """
+    entity(mark,root,1).
+    attribute((mark,type),1,tick).
+    entity(encoding,1,2).
+    attribute((encoding,channel),2,color).
+
```
domoritz, line 298: remove empty lines like this

draco/asp/hard.lp
```diff
@@ -43,7 +43,8 @@ violation(count_with_field) :-
 
 % @hard(point_tick_bar_without_x_or_y) Point, tick, and bar require x or y channel.
 violation(point_tick_bar_without_x_or_y) :-
-    entity(encoding,_,E),
```
domoritz, line 46: Ha, looks like we found a bug. Excellent.

draco/tests/test_hard.py
```diff
@@ -1,316 +1,313 @@
...
 
-    assert is_satisfiable(
-        p
-        + """
+    assert (
+        get_violations(
+            """
     attribute((mark,type),m1,text).
     entity(encoding,m1,e1).
     attribute((encoding,channel),e1,text).
```
domoritz, line 12: remove this empty line and others like it

draco/tests/test_hard.py
```diff
@@ -1,316 +1,313 @@
...
 
-    :- violation(_).
     """
+        )
+        == ["encoding_no_field_and_not_count"]
```
domoritz, line 15: Hmm, I think I would prefer not having any violations in the passing examples. What do you think?

domoritz (changes requested):

draco/tests/test_hard.py
JunranY: I was thinking that when we add more hard constraints, new violations could pop up. I was not sure whether I should try to keep some passing examples or add violations when that happens?

domoritz: Hmm good points. 

What I was thinking is that we want our tests to ensure that the constraints are written correctly, which the old format did. We should also ensure that constraints are not contained within each other because otherwise some constraints are unnecessary. However, the new format here makes it harder to test only the thing that matters. 

Let's leave this for now. I'll think more about this pull request but for now don't work on it.

JunranY: Got it. I think the old format might work better for this purpose by isolating the constraints.
But on the other hand, the new format does help with catching bugs that I wasn't aware of.

domoritz: Maybe we need separate tests. Can you send a pull request to fix the problem you found?

JunranY: Oh I mean the one above, so it's already taken care of.

PR #191, opened 2022-01-24
feat: port more hard constraints from Draco 1

JunranY:
I kept the old test format for test_hard for now. 

I'm unsure about some constraints and left them out. The questions I had:
- How do I define and use zero(E)
- Does the numbers of bin have to be specified like attribute((encoding, binning),E,10)? One example from documentation used “binning:true” (ie. attribute((encoding, binning),E)).
- How to define extent?


draco/asp/hard.lp
```diff
@@ -1,24 +1,98 @@
 % ====== Hard constraints ======
 
-% @hard(text_mark_without_text_channel) Text mark requires text encoding.
-violation(text_mark_without_text_channel) :-
-    attribute((mark,type),M,text),
-    not attribute(has_channel,M,text).
+% === Within Encodings ===
```
domoritz, line 3: I don't think these annotations work anymore with the new system, do they? Maybe we should split the hard constraints into multiple files.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,73 @@
 attribute(has_channel,M,C) :-
     entity(encoding,M,E),
     attribute((encoding,channel),E,C).
+
+% @helper(check_field_type) Whether a field is of type T.
+attribute(check_field_type,F,T) :-
```
domoritz, line 9: Since we are not checking, I think we should rename this helper

```suggestion
attribute(field_type,F,T) :-
```

draco/asp/helpers.lp
```diff
@@ -4,3 +4,73 @@
...
+    attribute((field,name),E,F),
```
domoritz, line 10:
```suggestion
    attribute((field,name),F,N),
```

The field should be F, not E (which we use for encoding).

draco/asp/helpers.lp
```diff
@@ -4,3 +4,73 @@
...
+    attribute((field,min),F,MIN),
+    attribute((field,max),F,MAX).
+
+% @helper(discrete) Whether the channel is discrete.
+attribute(discrete,M,C) :-
+    entity(mark,V,M),
+    entity(scale,V,E),
+    attribute((scale,channel),E,C),
+    attribute((scale,type),E,(ordinal;categorical)).
+attribute(continuous,M,C) :-
```
domoritz, line 46: Should this be under continuous?

draco/tests/test_hard.py
```diff
@@ -296,20 +287,739 @@ def test_point_tick_bar_without_x_or_y():
...
+    attribute((scale,type),5,categorical).
+
+    :- violation(_).
+    """
+    )
+
+    assert not is_satisfiable(
+        p
+        + """
+    entity(view,root,1).
```
domoritz, line 939: Uff, this becomes hard to read. Can we make a smaller example or use the nested representation?

draco/tests/test_hard.py
```diff
@@ -296,20 +287,739 @@ def test_point_tick_bar_without_x_or_y():
...
+        + """
+    attribute((scale,channel),4,size).
+    attribute((scale,type),4,categorical).
+
+    :- violation(_).
+    """
+    )
+
+
+# def test_size_negative():
```
domoritz, line 757: Commented out code

draco/tests/test_hard.py
domoritz: Same for some of the specs above. Use your judgment.

domoritz (changes requested):

domoritz:
> How do I define and use zero(E)

Zero is also a scale property, I think. 

> Does the numbers of bin have to be specified like attribute((encoding, binning),E,10)? One example from documentation used “binning:true” (ie. attribute((encoding, binning),E)).

Let's use numbers. Please fix the docs. 

> How to define extent?

We could define the scale domain and ranges explicitly on a scale. It would be a property with two values.

draco/asp/hard.lp
JunranY: I thought these are just comments. Should I delete them instead?

domoritz: They made sense in Draco 1 where we didn't parse the file as we do now. If you look at the parsed `Block`s, you will see how things are parsed with this comment. 

We would need to add support for some annotation to the parser. We could for example add a new property `group` to the `Block` and then set it based on some special comment. 

For now, I would say remove it.

domoritz:
Please update after https://github.com/cmudig/draco2/pull/194 is merged

draco/asp/helpers.lp
```diff
@@ -4,3 +4,55 @@
...
+    attribute((scale,type),E,T).
+
+% @helper(check_channel_field) Whether the channel is referring to the field.
+attribute(channel_field,M,C,N) :-
+    entity(encoding,M,E),
+    attribute((encoding,channel),E,C),
+    attribute((encoding,field),E,N).
+
+% @helper(extent) extent of a field.
+attribute(extent,N,MIN,MAX) :-
```
domoritz, line 36: Field should probably be F

draco/asp/helpers.lp
```diff
@@ -4,3 +4,55 @@
...
+attribute(discrete,M,C) :-
+    attribute(mark_scale_channel,M,E,C),
+    attribute((scale,type),E,(ordinal;categorical)).
+
+% @helper(continuous) Whether the channel is continuous.
+attribute(continuous,M,C) :-
+    attribute(mark_scale_channel,M,E,C),
+    attribute((scale,type),E,(linear;log)).
+
+% @helper(mark_x_y_channels) whether a mark has both x and y channels.
```
domoritz, line 50:
```suggestion
% @helper(mark_x_y_channels) Whether a mark has both x and y channels.
```

draco/asp/hard.lp
```diff
@@ -39,8 +117,35 @@ violation(count_with_field) :-
...
 % @hard(point_tick_bar_without_x_or_y) Point, tick, and bar require x or y channel.
 violation(point_tick_bar_without_x_or_y) :-
     attribute((mark,type),M,(point;tick;bar)),
     not attribute(mark_channel,M,x),
     not attribute(mark_channel,M,y).
+
+% @hard(line_area_without_x_y) Line and area require x and y channel.
+violation(line_area_without_x_y) :-
+    attribute((mark,type),M,(line;area)),
+    not attribute(mark_x_y_channels,M).
```
domoritz, line 139: Can you write this without the helper? I think you can use either two constraints or as a shortcut use `;`. I'd like to avoid helpers that are easily replaced.

draco/asp/hard.lp
```diff
@@ -39,8 +117,35 @@ violation(count_with_field) :-
...
+% @hard(line_area_with_discrete) Line and area cannot have two discrete(ie. x and y cannot both be discrete).
+violation(line_area_with_discrete) :-
+    attribute((mark,type),M,(line;area)),
+    attribute(discrete,M,x),
+    attribute(discrete,M,y).
+
+% @hard(bar_tick_continuous_x_y) Bar and tick cannot have both x and y continuous.
+violation(bar_tick_continuous_x_y) :-
+    attribute((mark,type),M,(tick;bar)),
+    attribute(mark_cont_channel,M,x),
```
domoritz, line 150: This helper name is very different from `discrete` above, although it is similar.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,55 @@
 attribute(mark_channel,M,C) :-
     entity(encoding,M,E),
     attribute((encoding,channel),E,C).
+
+% @helper(mark_scale_channel) Whether a mark has an scale for a channel.
+attribute(mark_scale_channel,M,E,C) :-
+    entity(mark,V,M),
+    entity(scale,V,E),
```
domoritz, line 11: Use `S` for scale.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,55 @@
...
+    attribute((scale,channel),E,C).
```
domoritz, line 12: Hmm, is this correct? Isn't E the scale?

draco/asp/hard.lp
```diff
@@ -39,8 +117,35 @@ violation(count_with_field) :-
...
     attribute((mark,type),M,(point;tick;bar)),
     not attribute(mark_channel,M,x),
     not attribute(mark_channel,M,y).
+
+% @hard(line_area_without_x_y) Line and area require x and y channel.
+violation(line_area_without_x_y) :-
+    attribute((mark,type),M,(line;area)),
+    not attribute(mark_x_y_channels,M).
+
+% @hard(line_area_with_discrete) Line and area cannot have two discrete(ie. x and y cannot both be discrete).
```
domoritz, line 141:
```suggestion
% @hard(line_area_with_discrete) Line and area cannot have two discrete positional encodings (i.e. x and y cannot both be discrete).
```

draco/asp/hard.lp
domoritz: I wonder whether it's cleaner to write constraints over scale types. Write a helper to add the scale type to an encoding and then ensure that the scale type is continuous. For the later, you can have a fact like `domain(continuous,(linear;log))`.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+    attribute((encoding,aggregate),E,count).
+violation(count_q_without_field) :-
+    entity(encoding,M,E),
+    attribute((encoding,aggregate),E,count),
+    attribute((encoding,channel),E,C),
+    attribute(scale_type,M,C,T),
+    T != linear,
+    T != log.
+
+% @hard(shape_discrete_non_ordered) Shape requires discrete and not ordered (nominal). Using ordinal would't make a difference in Vega-Lite.
```
domoritz, line 74: Update description. We should not refer to Vega-Lite here since we want to be agnostic. 

```suggestion
% @hard(shape_discrete_non_ordered) Shape requires discrete and not ordered (nominal).
```

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+
+% @hard(count_q_without_field) Count has to be quantitative and not use a field.
+violation(count_q_without_field) :-
+    attribute((encoding,field),E,_),
+    attribute((encoding,aggregate),E,count).
+violation(count_q_without_field) :-
+    entity(encoding,M,E),
+    attribute((encoding,aggregate),E,count),
+    attribute((encoding,channel),E,C),
+    attribute(scale_type,M,C,T),
```
domoritz, line 70: This can be simplified I think. Let's add an attribute (automatically inferred from the scale type on the mark) to encodings for the scale type.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+    attribute((encoding,channel),E,C),
+    attribute((encoding,aggregate),E,_),
+    attribute(scale_type,M,C,categorical).
+
+% @hard(aggregate_detail) Detail cannot be aggregated.
+violation(aggregate_detail) :-
+    attribute((encoding,channel),E,detail),
+    attribute((encoding,aggregate),E,_).
+
+% @hard(count_q_without_field) Count has to be quantitative and not use a field.
```
domoritz, line 62: Let's not talk about quantitative encodings. The scale has to be quantitative. 

I think we can split this one constraint into two.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+
+% @hard(aggregate_t_valid) Temporal only supports min and max.
+violation(aggregate_t_valid) :-
+    attribute(field_type,N,datetime),
+    attribute((encoding,field),E,N),
+    attribute((encoding,aggregate),E,A),
+    A != min,
+    A != max.
+
+% @hard(aggregate_nominal) Cannot aggregate categorical.
```
domoritz, line 50: We also cannot aggregate ordinal.

domoritz: So let's name this `aggregate_categorical`.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+    not attribute((scale,type),E,categorical).
+
+% @hard(size_nominal)  Size implies order so nominal is misleading.
+violation(size_nominal) :-
+    attribute((scale,channel),E,size),
+    attribute((scale,type),E,categorical).
+
+% @hard(size_negative) Do not use size when data is negative as size implies that data is positive.
+violation(size_negative) :-
+    attribute(channel_field,_,size,N),
```
domoritz, line 91: This is hard to read. Why does `channel_field` have four components? It's unclear why we have `_` here. 

Note that creating unnecessary facts (especially ones with many components) can explode in space since the grounder has to instantiate many options.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+    attribute((encoding,aggregate),E,count),
+    attribute((encoding,channel),E,C),
+    attribute(scale_type,M,C,T),
+    T != linear,
+    T != log.
+
+% @hard(shape_discrete_non_ordered) Shape requires discrete and not ordered (nominal). Using ordinal would't make a difference in Vega-Lite.
+violation(shape_discrete_non_ordered) :-
+    attribute((scale,channel),E,shape),
+    not attribute((scale,type),E,categorical).
```
domoritz, line 77: Categorical is reserved for color. Let's add a constraint for that and also clean up the constraints here.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
-% @hard(no_encodings) There has to be at least one encoding for every mark.
-violation(no_encodings) :-
-    entity(mark,_,M),
-    not entity(encoding,M,_).
+% @hard(aggregate_o_valid) Oridnal only supports min, max, and median.
+violation(aggregate_o_valid) :-
+    entity(encoding,M,E),
+    attribute((encoding,channel),E,C),
+    attribute((encoding,aggregate),E,A),
+    attribute(scale_type,M,C,ordinal),
```
domoritz, line 37: Categorical would also work here, actually.

draco/asp/hard.lp
```diff
@@ -1,24 +1,97 @@
...
+    attribute(field_type,N,(string;boolean)),
+    attribute(scale_type,M,C,T),
+    attribute(channel_field,M,C,N),
+    T != ordinal.
 
-% @hard(text_channel_without_text_mark) Text channel requires text mark.
-violation(text_channel_without_text_mark) :-
-    attribute(mark_channel,M,text),
-    not attribute((mark,type),M,text).
+% @hard(bin_q_o) Can only bin quantitative or ordinal.
```
domoritz, line 10: Hmm, I'm not ready for binning yet. How binning relates to scales is something we need to think about a bit more. Let's remove binning constraints from the pull request for now.

domoritz (changes requested):
Thank you for the constraints. I think we can refactor them a bit. 

Could you write down how you think about scale types and how they relate to marks and data types? We need a clear understanding to write precise constraints. 

After this pull request, we could start looking into doing some actual recommendation. Then we can add more constraints as we see that the generation step creates invalid visualizations.

domoritz:
If say we could add two pieces on information to encodings. First, the scale details so we can reason about things like domains etc. Second, the type of encoding, which for now could be continuous (linear, log) or discrete (ordinal, categorical, or binned).

draco/asp/hard.lp
JunranY: removed. Similar to ordinal, we can aggregate categorical with min, max and median.

draco/asp/hard.lp
JunranY: removed.

draco/asp/hard.lp
JunranY: split into `count_without_field` and `count_q`

draco/asp/hard.lp
JunranY: replaced with `categorical_not_color`

JunranY:
> Could you write down how you think about scale types and how they relate to marks and data types? We need a clear understanding to write precise constraints.
Here is what I think: 
- Scale types reflect the result from mark encoding for data fields. So there are two sets of relationship/constraints 1) between field type and encoding, and 2) encoding and scale type. Encoding is like a transition between field types and scale types.  Maybe we can refer to this table form Vega-lite:
By default, Vega- Lite use the fotowing scale types for the following data types and encoding channels
- Also need to make sure each encoding channel has a corresponding scale channel

I'm not 100% sure about if I got the above two unclasped comments correct. Thanks for the review! 
Also, I think the linter didn't auto-fix the format.

domoritz:
> Also, I suppose the syntax for field need to be modified in [examples](https://dig.cmu.edu/draco2/facts/examples.html#data-for-the-examples) and I should refer to [this](https://dig.cmu.edu/draco2/facts/schema.html#field-properties)

I think the example is right but could be simplified by using names such as `f1`, `f2` etc and adding a `(field,name)` property. I addressed this in https://github.com/cmudig/draco2/pull/200.

docs/facts/examples.ipynb
```diff
@@ -279,6 +279,31 @@
    "id": "3b1b4c39",
    "metadata": {},
    "outputs": [
+    {
+     "name": "stdout",
```
domoritz, line 283: I think you might need to rerun the book command. Run `make` and then commit the chnages.

.github/workflows/test.yml
```diff
@@ -40,5 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
-      - uses: psf/black@21.12b0
+      - uses: psf/black@stable
```
domoritz, line 43: If you merge main, this should disappear from the diff.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,36 @@
...
+    entity(scale,R,S),
+    attribute((scale,channel),S,C).
+
+% @helper(extent) extent of a field.
+attribute(extent,F,MIN,MAX) :-
+    attribute((field,min),F,MIN),
+    attribute((field,max),F,MAX).
+
+% @helper(discrete) Whether the channel is discrete.
+attribute(discrete,M,C) :-
```
domoritz, line 25: replace with `attribute(channel_type,V,(continuous;discrete))`. Note that V is a view, because I think the type is the same for the view and not just a mark. We can add another helper for a mark but I wonder whether it's better to write the constraints over the view.

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
+    attribute((encoding,field),E,F),
+    attribute((encoding,type),E,(linear;log)).
 
-% @hard(text_channel_without_text_mark) Text channel requires text mark.
-violation(text_channel_without_text_mark) :-
-    attribute(mark_channel,M,text),
-    not attribute((mark,type),M,text).
+% @hard(log_non_positive) Cannot use log if the data is negative or zero.
+violation(log_non_positive) :-
+    attribute(extent,F,MIN,_),
```
domoritz, line 11: We don't need the helper here, no? We cold get MIN directly.

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
 % @hard(bin_and_aggregate) Cannot bin and aggregate.
 violation(bin_and_aggregate) :-
     attribute((encoding,binning),E,_),
     attribute((encoding,aggregate),E,_).
 
-% @hard(no_encodings) There has to be at least one encoding for every mark.
-violation(no_encodings) :-
-    entity(mark,_,M),
-    not entity(encoding,M,_).
+% @hard(aggregate_o_valid) Oridnal only supports min, max, and median.
```
domoritz, line 21: Typo.

Also, not what the actual constraint is. 

I would say categorical also does not really support min, max, and median.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,36 @@
...
+    attribute(mark_scale_channel,M,S,C),
+    attribute((scale,type),S,(ordinal;categorical)).
+
+% @helper(continuous) Whether the channel is continuous.
+attribute(continuous,M,C) :-
+    attribute(mark_scale_channel,M,S,C),
+    attribute((scale,type),S,(linear;log)).
+
+% @helper(encoding_type)
+attribute((encoding,type),E,T) :-
```
domoritz, line 35:
```suggestion
attribute((encoding,scale_type),E,T) :-
```

Otherwise, it's not clear what type we are talking about.

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
+violation(aggregate_o_valid) :-
+    attribute((encoding,aggregate),E,A),
+    attribute((encoding,type),E,(ordinal;categorical)),
+    A != min,
+    A != max,
+    A != median.
+
+% @hard(aggregate_t_valid) Temporal only supports min and max.
```
domoritz, line 29:
```suggestion
% @hard(aggregate_t_valid) Temporal scale only supports min and max.
```

draco/asp/hard.lp
domoritz:
```suggestion
% @hard(aggregate_o_valid) Ordinal scale only supports min, max, and median.
```

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
+violation(aggregate_detail) :-
+    attribute((encoding,channel),E,detail),
+    attribute((encoding,aggregate),E,_).
+
+% @hard(count_without_field) Count cannot use a field.
+violation(count_without_field) :-
+    attribute((encoding,field),E,_),
+    attribute((encoding,aggregate),E,count).
+
+% @hard(count_without_q) Count has to have a quantitative scale.
```
domoritz, line 47:
```suggestion
% @hard(count_without_q) Count has to have a continuous scale.
```

Look at https://github.com/d3/d3-scale#api-reference for inspiration here. I think calling it continuous is clearer.

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
+violation(count_without_q) :-
+    attribute((encoding,aggregate),E,count),
+    attribute((encoding,type),E,T),
+    T != log,
+    T != linear.
```
domoritz, line 52: We should probably add some fact for all the continuous scales. Make it similar to `non_positional` etc that we use for channel.

draco/asp/hard.lp
```diff
@@ -1,24 +1,68 @@
...
+
+% @hard(categorical_not_color) Categorical only applied to color channel.
```
domoritz, line 54:
```suggestion
% @hard(categorical_not_color) Categorical only works with color channel.
```

draco/asp/hard.lp
```diff
@@ -39,8 +88,35 @@ violation(count_with_field) :-
...
+violation(line_area_with_discrete) :-
+    attribute((mark,type),M,(line;area)),
+    attribute(discrete,M,x),
+    attribute(discrete,M,y).
+
+% @hard(bar_tick_continuous_x_y) Bar and tick cannot have both x and y continuous.
+violation(bar_tick_continuous_x_y) :-
+    attribute((mark,type),M,(tick;bar)),
+    attribute(continuous,M,x),
+    attribute(continuous,M,y).
```
domoritz, line 122:
```suggestion
    attribute(continuous,M,y).

```

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+        + """
+    attribute((field,type),temperature,number).
+    attribute((field,type),name,string).
+    entity(mark,root,1).
+    entity(encoding,1,3).
+    entity(scale,root,5).
+    attribute((encoding,channel),3,x).
+    attribute((encoding,field),3,temperature).
+    attribute((scale,channel),5,x).
+    attribute((scale,type),5,linear).
```
domoritz, line 340: Make things a bit more readable by adding space and grouping related facts together. Apply this principle below as well. 

```suggestion
    attribute((field,type),temperature,number).
    attribute((field,type),name,string).
    
    entity(mark,root,1).
    entity(encoding,1,3).
    attribute((encoding,channel),3,x).
    attribute((encoding,field),3,temperature).
    
    entity(scale,root,5).
    attribute((scale,channel),5,x).
    attribute((scale,type),5,linear).
```

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+    """
+    )
+
+    assert (
```
domoritz, line 344: Add a comment what is wrong here. It's really hard to parse otherwise.

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+    attribute((encoding,channel),3,x).
+    attribute((encoding,field),3,temperature).
+    attribute((scale,channel),5,x).
+    attribute((scale,type),5,linear).
+    """
+        )
+        == ["enc_type_valid"]
+    )
+
+    # shared scale
```
domoritz, line 361: Unclear why it's invalid if we share a scale.

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+    entity(encoding,1,2).
+    entity(scale,0,4).
+    attribute((encoding,channel),2,x).
+    attribute((encoding,field),2,precipitation).
+    attribute((scale,channel),4,x).
+    attribute((scale,type),4,linear).
+    """
+    )
+
+    assert (
```
domoritz, line 401: add a test case where we have a log scale with only positive numbers

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+                }
+            )
+        )
+    )
+
+    # multiple views
+    assert (
+        list_violations(
+            b.program
+            + ("\n").join(
```
domoritz, line 805: Since `run_clingo` actually supports taking an iterator of strings instead of a string, I wonder whether we can write `list_violations` also in a way where it supports an iterator.  Then we could write

```py
list_violations([b.program] + dict_to_facts(...))
```

draco/tests/test_hard.py
```diff
@@ -319,3 +320,596 @@ def test_point_tick_bar_without_x_or_y():
...
+                                    {"channel": "x", "type": "linear"},
+                                    {"channel": "y", "type": "linear"},
+                                ],
+                            },
+                        ]
+                    }
+                )
+            )
+        )
+        == ["bar_tick_continuous_x_y"]
```
domoritz, line 840: It's hard to see the error. Maybe add a comment where something is wrong similar to https://github.com/cmudig/draco2/blob/b6fba8c31bd2377a617b4a3aaf186734f9229607/draco/tests/test_spec_utils.py#L25

domoritz (changes requested):

draco/asp/hard.lp
JunranY: added in define.lp

draco/asp/hard.lp
JunranY: I did include categorical.

I also think this constraint is too indirect. Maybe this one should be about aggregation and data type, rather than the resulted scale?

draco/asp/helpers.lp
JunranY: I'm a bit confused here. A scale has multiple channels, right? And the scales can be written in different levels(root/view)

draco/tests/test_hard.py
JunranY: same with the previous case. only the structure is different(with/without a shared scale)

draco/asp/hard.lp
JunranY: added two new constraints for aggregate and bin related to data types.

deleted `aggregate_o_valid` but we probably need a new version in the future.

JunranY:
made all changes except for https://github.com/cmudig/draco2/pull/191#discussion_r798164766

draco/asp/helpers.lp
domoritz: A scale always belongs to one channel. A scale is defined on a view but if there is only one view, we can make it implicit and define everything on the root. At least that's how I imagined it until now but I'm happy to consider alternatives.

JunranY: I mean a view can have multiple scales, one for each channel. 

By "different levels", I mean in the dict format a scale can be at the same level of the mark, but if there are multiple views where the marks share a scale, then the scale can be written at the same level as the views. 

This helper was written for the `line_area_with_discrete`(x and y channels cannot both be discrete) constraint. 
I think this constraint indicates:
 1) line and area should have both x and y encoding channels 
 2) each channel should have a corresponding scale 
 3) x and y scales cannot be both discrete. 

Maybe I should break this into 3 constraints, and the second one can be shared by similar constraints like `bar_tick_continuous_x_y` that relating marks and scales?

domoritz: I think the separate constraints make sense. 

For discrete vs continuous, we need to be a bit careful. A channel may have a linear scale but still be discrete if e.g. the channel is binned. We could get around this issue by a) adding new scale types (e.g. bin-linear) or b) having an encoding type that is discrete/continuous and that needs to be consistent with the scale type and possible transformations.

draco/asp/hard.lp
```diff
@@ -1,24 +1,74 @@
 % ====== Hard constraints ======
 
-% @hard(text_mark_without_text_channel) Text mark requires text encoding.
-violation(text_mark_without_text_channel) :-
-    attribute((mark,type),M,text),
-    not attribute(mark_channel,M,text).
+% @hard(enc_type_valid) Primitive type has to support data type.
```
domoritz, line 3:
```suggestion
% @hard(scale_type_data_type) Primitive type has to support scale type.
```

domoritz (changes requested):

draco/asp/hard.lp
```diff
@@ -1,24 +1,75 @@
...
 
-% @hard(text_mark_without_text_channel) Text mark requires text encoding.
-violation(text_mark_without_text_channel) :-
-    attribute((mark,type),M,text),
-    not attribute(mark_channel,M,text).
+% @hard(scale_type_data_type) Primitive type has to support scale type.
+violation(scale_type_data_type) :-
+    attribute((field,type),F,(string;boolean)),
+    attribute((encoding,field),E,F),
+    attribute((encoding,scale_type),E,(linear;log)).
```
domoritz, line 7: Please use `domain(continuous_scale,T)` here.

draco/asp/helpers.lp
```diff
@@ -4,3 +4,21 @@
 attribute(mark_channel,M,C) :-
     entity(encoding,M,E),
     attribute((encoding,channel),E,C).
+
+% @helper(mark_scale_channel) Whether a mark has a scale for a channel.
```
domoritz, line 8: I think this helper is not quite correct. We could have a scale defined on the view and the mark. This rule implies that we would have two scales for the encoding. We should only use the closest scale definition to the encoding.

draco/asp/hard.lp
domoritz, line 7: Please use `domain(continuous_scale,T)` here.

domoritz (changes requested):

draco/asp/helpers.lp
```diff
@@ -4,3 +4,21 @@
...
+attribute(mark_scale_channel,M,S,C) :-
```
domoritz, line 9: This can probably be `attribute((mark, scale_type),M,T)` since we don't need the full scale in marks right now. What is the reason that you decided for `mark_scale_channel`?

domoritz (approved):
Looks good. In a follow up, please add a constraint that ensures that we don't have the same scale defined in a view and a mark.

JunranY:
> Looks good. In a follow up, please add a constraint that ensures that we don't have the same scale defined in a view and a mark.

done in: https://github.com/cmudig/draco2/blob/b1b1672fc8b4cf97f82e00e8f5d86d0c8232b5f1/draco/asp/hard.lp#L137

PR #209, opened 2022-02-11
refactor: simplify mark scale channel

domoritz:


JunranY (approved):

domoritz:
> I think writing the helper in this way requires another constraint: cannot define multiple channels under the same encoding entity or the same scale entity, which seems like it's made for draco facts structure.

I'm not sure I fully follow this. I agree that we need those constraints but that's outside the scope of this refactoring, right? 

What do you mean by " which seems like it's made for draco facts structure."?

JunranY:
Yes, it's outside of this pr. I was keeping a note of what should be added. 

> What do you mean by " which seems like it's made for draco facts structure."?

I mean i think it serves a purpose similar to those in constriants.lp?

PR #214, opened 2022-02-16
fix: reduce grounding for repeat_channel

JunranY:
by leveraging asymmetry.

I did try using `--text` and saw the difference between "!=" and ">".

Is there a way to check groundings within the tests? 
I tried to search it in clingo python API, but couldn't find how to do it.


JunranY:
not sure why this pr contains old commits :sweat:

domoritz:
Great. What's the difference in number of lines in the grounded program?

JunranY:
85 lines.

PR #224, opened 2022-02-24
fix: make_fact argument type

JunranY:
trying to fix the test


draco/fact_utils.py
```diff
@@ -34,7 +34,7 @@ def stringify(value):
     return value[:1].lower() + value[1:]
 
 
-def make_fact(kind: FactKind, values=List) -> str:
+def make_fact(kind: FactKind, values: Tuple = ()) -> str:
     """Create an ASP fact from a list of values. The function generates either
```
domoritz, line 38: The comment says list. Should we use `Iterable`?

domoritz (changes requested):

PR #233, opened 2022-03-01
feat: add hard constraints

JunranY:
included binned encodings in discrete scales
added zero property in scale


JunranY:
have a few things to fix before the review

docs/facts/examples.ipynb
```diff
@@ -340,39 +340,49 @@
...
-       " 'attribute((scale,type),5,linear).']"
-      ]
-     },
-     "execution_count": 37,
-     "metadata": {},
-     "output_type": "execute_result"
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "<block>:324:5-30: info: atom does not occur in any rule head:\n",
```
domoritz, line 355: This is an error.

draco/asp/hard.lp
```diff
@@ -192,4 +191,48 @@ violation(aggregate_not_all_continuous):-
...
+    T != number.
+
+% @hard(bar_area_without_zero) Bar and area mark requires scale of continuous to start at zero.
+violation(bar_area_without_zero) :-
+    attribute((mark,type),M,(bar;area)),
+    entity(encoding,M,E),
+    attribute((encoding,channel),E,x),
+    not attribute((encoding,binning),E,_),
+    attribute((mark,scale),M,S),
+    attribute((scale,channel),S,x),
```
domoritz, line 225: can we write `x;y`?

draco/asp/helpers.lp
```diff
@@ -24,8 +26,28 @@ attribute((encoding,scale_type),E,T) :-
...
 
-% @helper(mark_channel_field) Whether a mark has an encoding for a field  and a channel.
+% @helper(mark_channel_field) Whether a mark has an encoding for a field and a channel.
 attribute(mark_channel_field,M,C,F) :-
     entity(encoding,M,E),
     attribute((encoding,channel),E,C),
     attribute((encoding,field),E,F).
+
+% @helper(mark_channel_scale_not_binned) Whether a mark has a scale type for a channel that is not binned.
+attribute(mark_channel_scale_not_binned,M,T,C) :-
```
domoritz, line 36: This seems like an odd helper. Why do we have a helper where you only get the scale when the encoding is not binned?

draco/asp/hard.lp
JunranY: I don't think so, the channel need to be exact the same across encoding and scale

draco/asp/helpers.lp
JunranY: I was trying to write the constraint for `bar_tick_continuous_x_y` (Bar and tick cannot have both x and y continuous.)
I think the prohibition for being continuous can be relaxed when the encoding is binned.

draco/asp/hard.lp
domoritz: Oh I missed that. Could you use a variable and then say that the variable is x;y?

draco/asp/helpers.lp
domoritz: Wouldn't it make more sense to call it `attribute(mark_channel_scale,M,T,C,B)` where B is true or false.

JunranY: I made some changes, but wasn't sure about the syntax.

domoritz: That looks great, don't you think?

JunranY: ohh I mean I wasn't sure the changes I made was the right way or not.

domoritz: There is no absolute right or wrong but it looks reasonable to me.

Issue #221, opened 2022-02-22
Add documentation for topK argument


domoritz:
https://github.com/cmudig/draco2/blob/9cd0494b4a52f7bea2dbb18bf3b80cd2b49e7ce1/draco/run.py#L33

PR #234, opened 2022-03-01
docs: document topK argument

Zehua-Zeng:
Issue #221


draco/run.py
```diff
@@ -37,7 +37,9 @@ def run_clingo(
 
     :param program: Program as a string or iterable of strings that will be
         concatenated.
-    :param models: Number of models to generate, defaults to 0 (meaning all models)
+    :param models: Number of models to generate, defaults to 0 (meaning all models).
+    :param topK: Whether returning the top K models. If false (default), the program
```
domoritz, line 41:
```suggestion
    :param topK: Whether to return the top K models. If false (default), the program
```

domoritz (approved):

draco/run.py
Zehua-Zeng: done

PR #236, opened 2022-03-03
docs: encoding field in examples refer to id rather than name

JunranY:
explicitly adding id worked.
the error in binned histogram is expected to be fixed with pr #233


docs/facts/examples.ipynb
```diff
@@ -326,53 +326,38 @@
...
-      ]
-     },
-     "execution_count": 37,
-     "metadata": {},
-     "output_type": "execute_result"
+     "ename": "AssertionError",
+     "evalue": "Invalid spec",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[0;31m---------------------------------------------------------------------------\u001b[0m",
```
domoritz, line 355: Fix this error

domoritz (changes requested):

PR #248, opened 2022-03-14
build: cache poetry builds for github actions

cabreraalex:


.github/workflows/book.yml
```diff
@@ -17,8 +17,21 @@ jobs:
...
+        uses: actions/cache@v2
+        id: cache
+        with:
+          path: ~/.virtualenvs
+          key: poetry-${{ hashFiles('**/poetry.lock') }}
+          restore-keys: |
+            poetry-${{ hashFiles('**/poetry.lock') }}
+      - name: Set Poetry config
+        run: |
+          poetry config virtualenvs.in-project false
```
domoritz, line 30: Isn't that the default? https://python-poetry.org/docs/configuration/#virtualenvsin-project

domoritz:
Awesome. So much faster.

PR #249, opened 2022-03-16
feat: change define.lp and add constraints to hard.lp for facets

Zehua-Zeng:


draco/asp/define.lp
```diff
@@ -47,3 +47,12 @@ domain((encoding,stack),(zero;center;normalize)).
 
 % @definition(scale_zero) Scale start with 0.
 domain((scale,zero),true).
+
+% @definition(facets) Definition for facets
+domain((facet,field),F) :- entity(field,_,F).
+
+domain(facet_channel,(row;col)).
+domain((facet,channel),C) :- domain(facet_channel,C).
```
JunranY, line 55:
```suggestion
domain((facet,channel),(row;col)).
```

draco/asp/hard.lp
```diff
@@ -230,6 +230,27 @@ violation(bar_area_without_zero) :-
     not attribute((scale,zero),S,_),
     C = (x;y).
 
+% @hard(row_or_col_c) Row and column require discrete. If number, need to be binned.
+violation(row_or_col_c) :-
```
JunranY, line 234: I think field with number type can be discrete too. For example if it's ordinal with a small cardinality.

draco/asp/define.lp
Zehua-Zeng: done

draco/asp/hard.lp
JunranY: @domoritz Do we want to write this one in soft constraints considering numeric field can be ordinal(but can't be explicitly indicated)?

domoritz: Yes, a numeric field can be discrete as well. We could have soft constraints that require high cardinality numbers to be binned to be discrete.

Zehua-Zeng: So we don't need this hard constraint to require numbers must be binned for faceting?

domoritz: I don't think we should have this as a hard constraint. The soft constraint should be a bit more thoughtful like "the cardinality of facets should be less than x" and then we should define the effective cardinality based on the data cardinality and binning. We also need to add a generate for binning. Then the silver will recommend binning when the cardinality is too high.

draco/asp/define.lp
```diff
@@ -47,3 +47,9 @@ domain((encoding,stack),(zero;center;normalize)).
 
 % @definition(scale_zero) Scale start with 0.
 domain((scale,zero),true).
+
+% @definition(facets) Definition for facets
+domain((facet,field),F) :- entity(field,_,F).
+domain((facet,channel),(row;col)).
+domain((facet,binning),(10;25;200)).
```
domoritz, line 54: Extract the binning options into something we can reuse for the encoding binning.

domoritz (approved):

draco/asp/define.lp
Zehua-Zeng: Do you mean to add a helper to know if a field is binned or not?

domoritz: No, add a fact with `binCounts()` or something so we can reuse the possible bin counts.

PR #257, opened 2022-03-29
feat: port the first soft constraint and test

JunranY:


draco/asp/soft.lp
```diff
@@ -0,0 +1,4 @@
+% ====== Preferences ======
+% @soft(aggregate) Prefer to use raw (no aggregate).
+violation(aggregate,E) :-
```
domoritz, line 3: This will not work entirely since we are forbidding violations right now. Should we add a list of violations that are forbidden or have a second type of violation that's for soft constraints? What do you think?

draco/tests/test_soft.py
```diff
@@ -0,0 +1,38 @@
...
+
+    assert (
+        list_violations(
+            b.program
+            + """
+    attribute((encoding,aggregate),e1,mean).
+    attribute((encoding,aggregate),e2,mean).
+    """
+        )
+        == ["aggregate"] * 2
```
domoritz, line 37: Great to count it twice. I wonder whether we should figure out a way to return the whole tuple, not just the name of the violation? This way we would know where the violation is happening. Could return for example `("aggregate", "e1")`.

domoritz (approved):

draco/asp/soft.lp
JunranY: Can we treat hard and soft equally, remove the forbidding violation requirement and assign hard with infinite weight?

domoritz: We discussed in a a meeting that while it's possible, it's not advisable.

PR #267, opened 2022-04-07
feat: add multi-view to generate.lp and clean it up; add argument for passing constants to run_clingo

JunranY:


draco/tests/test_generate.py
```diff
@@ -0,0 +1,27 @@
...
+    prog = (
+        generate.program
+        + define.program
+        + helpers.program
+        + hard.program
+        + constraints.program
+    )
+    with open(asp_path / "examples" / "scatter.lp") as file:
+        scatter = file.readlines()
+    scatter = ("").join(scatter)
```
domoritz, line 19:
```suggestion
        scatter = file.read()
```

draco/tests/test_generate.py
```diff
@@ -0,0 +1,27 @@
...
+
+    models = list(
+        run_clingo(
+            prog + scatter + ":- {entity(mark,_,_)} < 3.", 10, False, {"max_marks": 3}
+        )
+    )
+
+    assert len(models) > 0.0
```
domoritz, line 27:
```suggestion
    assert len(models) > 0
```

draco/asp/generate.lp
```diff
@@ -1,19 +1,27 @@
...
 { entity(view,root,(v,0)) } = 1.
 
-% maximum number of additional marks for each view
-#const max_extra_marks = 1.
-mark_id(0..max_extra_marks).
-{ entity(mark,V,(m,M)) : entity(view,_,V), mark_id(M) }.
-:- not entity(mark,_,(m,ID)), entity(mark,_,(m,ID-1)), mark_id(ID), mark_id(ID-1).
+% maximum number of non-layered views.
+#const max_views = 1.
+{ entity(view,root,(v,V)) } :- V = 0..(max_views-1).
```
domoritz, line 13: Does this work?

```suggestion
{ entity(view,root,(v,0..max_views-1)) }.
```

draco/asp/generate.lp
```diff
@@ -1,19 +1,27 @@
 % ====== Generators ======
 
+% helpers to generate attributes based on how they are required.
```
domoritz, line 3:
```suggestion
% helpers to generate attributes based on whether they are required.
```

draco/asp/generate.lp
```diff
@@ -1,19 +1,27 @@
...
+{ entity(mark,(v,V),(m,M)) } :- entity(view,_,(v,V)), N = 0..(max_marks-1), M = max_marks * V + N.
+:- entity(mark,(v,V),(m,ID)), not entity(mark,(v,V),(m,ID-1)), ID > 1 + max_marks * V.
 
 % maximum number for total encoding channels
-#const max_extra_encs = 4.
-encoding_id(M,E) :- entity(mark,_,(m,M)), N = 0..max_extra_encs, E = ((max_extra_encs + 1) * M) + N.
-{ entity(encoding,(m,M),(e,E)) : entity(mark,_,(m,M)), encoding_id(M,E) }.
-:- not entity(encoding,_,(e,ID)), entity(encoding,_,(e,ID-1)),encoding_id(_,(e,ID)), encoding_id(_,(e,ID-1)).
+#const max_encs = 4.
+{ entity(encoding,(m,M),(e,E)) } :- entity(mark,_,(m,M)), N = 0..(max_encs-1), E = max_encs * M + N.
```
domoritz, line 23: What's the purpose of `E = max_encs * M + N`?

draco/run.py
```diff
@@ -40,13 +43,25 @@ def run_clingo(
     :param models: Number of models to generate, defaults to 0 (meaning all models).
     :param topK: Whether to return the top K models. If false (default), the program
         will not optimize the output models.
+    :param constants: Dictionary of constants used in genenerator
+        for the maximum number of individual entity types (view, mark and encoding).
```
domoritz, line 47:
```suggestion
        for the maximum number of individual entity types (view, mark, and encoding).
```

draco/run.py
```diff
@@ -40,13 +43,25 @@ def run_clingo(
     :param models: Number of models to generate, defaults to 0 (meaning all models).
     :param topK: Whether to return the top K models. If false (default), the program
         will not optimize the output models.
+    :param constants: Dictionary of constants used in genenerator
```
domoritz, line 46: I wonder whether we should treat this as more general. What's happening is that we are passing in constants but how exactly they are used is irrelevant for `run`.

draco/run.py
```diff
@@ -40,13 +43,25 @@ def run_clingo(
...
+        the generator will refer to the partial specification.
     :yield: The models.
     """
     if not isinstance(program, str):
         program = "\n".join(program)
 
+    flags = []
+    if constants:
+        for key, value in constants.items():
+            if key in ["max_views", "max_marks", "max_encs"]:
```
domoritz, line 60: I think we should support any constant ant not just the ones we know about. Imagine if someone wants to run Draco with custom asp, then `run` should still what they use since they cannot modify the Python code when they import the Draco library.

draco/tests/test_generate.py
```diff
@@ -0,0 +1,27 @@
...
+        + hard.program
+        + constraints.program
+    )
+    with open(asp_path / "examples" / "scatter.lp") as file:
+        scatter = file.readlines()
+    scatter = ("").join(scatter)
+
+    models = list(
+        run_clingo(
+            prog + scatter + ":- {entity(mark,_,_)} < 3.", 10, False, {"max_marks": 3}
```
domoritz, line 23: Let's add another unit test (in the run tests) that tests whether you can pass in a constant. 

The program could be

```asp
#const foo.
a(0...foo).
```

And then you can test whether the number of `a`s is `foo` and vary `foo` a bit in the test.

domoritz (changes requested):

draco/asp/generate.lp
JunranY: It generates N encodings for each mark.
ex. If there are 2 marks: (m,0) and (m,1). N=4. 
Then we generate (e,0) to (e,3) for (m,0), and (e,4) to (e,7) for (m,1).
Although one problem is that since N is not enforced, there could be gaps in the ids, like (e,0) for (m,0), and (e,4),(e,5) for (m,1) such that (e,1..3) doesn't exist in the sequence. 
Is there a better way to avoid this?

domoritz: Hmm, interesting. I guess we could add the mark to the ID tuple so that we could have numbers that are consecutive and start from 0. So like (e,2,0) where 2 is the mark and 0 is the encoding id.

JunranY: ha. I asked about this two weeks ago. Since we are having multiple views, things could become like (v,0), (m,0,0), (e,0,0,0)

domoritz: I think this could work. The other option is instead to limit the number of new encodings overall (not per view). That could also work but may in the end work less well. Not sure.

JunranY: done

domoritz (approved):

PR #268, opened 2022-04-07
feat: hard constraints for stacking

Zehua-Zeng:


draco/asp/hard.lp
```diff
@@ -265,6 +265,84 @@ violation(col_no_x) :-
...
+    attribute((mark,type),M,(bar;area)),
+    not attribute(mark_with_stack,M).
+
+% @hard(stack_without_discrete_color_or_detail) Can only use stack if we also use discrete color, or detail.
+violation(stack_without_discrete_color_or_detail) :-
+    attribute(mark_with_stack,M),
+    not attribute(mark_channel_discrete_or_binned,M,color),
+    not attribute((mark,channel),M,detail).
+
+% @hard(stack_detail_without_q_color) If we use stack and detail, we also have to use quantitative color.
```
JunranY, line 293: not sure what `stack_detail_without_q_color` and `stack_detail_without_q_color_agg` are about.
The description says about quantitative, but it's not reflected in the constraints.

draco/asp/hard.lp
```diff
@@ -265,6 +265,84 @@ violation(col_no_x) :-
...
+% @hard(stack_with_non_positional_non_agg) Cannot use non positional continuous with stack unless it's aggregated.
+violation(stack_with_non_positional_non_agg) :-
+    attribute(mark_with_stack,M),
+    entity(encoding,M,E),
+    attribute((encoding,channel),E,C),
+    domain(non_positional,C),
+    not attribute((encoding,aggregate),E,_),
+    not attribute(mark_channel_discrete_or_binned,M,C).
+
+% @hard(shape_with_cardinality_gt_eight) Vega-Lite currently supports 8 shapes.
```
JunranY, line 328: We may want to remove this for things other than Vega-Lite

draco/asp/hard.lp
```diff
@@ -265,6 +265,84 @@ violation(col_no_x) :-
...
+    attribute((encoding,channel),C,shape),
+    attribute((encoding,field),C,F),
+    U > 8.
+
+% @hard(color_with_cardinality_gt_twenty) At most 20 categorical colors.
+violation(color_with_cardinality_gt_twenty) :-
+    entity(field,root,F),
+    attribute((field,unique),F,U),
+    attribute((encoding,channel),E,color),
+    attribute((encoding,field),C,F),
```
JunranY, line 341: do you mean:
```suggestion
    attribute((encoding,field),E,F),
```
Also need to ensure scale and encoding are for the same view

draco/tests/test_hard.py
```diff
@@ -2008,6 +2008,631 @@ def test_col_no_x():
     )
 
 
+def test_stack_without_bar_area():
```
JunranY, line 2011: add more comment to describe the test cases and break each example into segments to be more readable.

draco/tests/test_hard.py
```diff
@@ -2008,6 +2008,631 @@ def test_col_no_x():
...
+    entity(scale,0,7).
+    attribute((scale,channel),7,color).
+    attribute((scale,type),7,categorical).
+    """
+    )
+
+    assert no_violations(
+        p
+        + """
+    entity(mark,0,1).
```
JunranY, line 2195: maybe we can try dict format with such long examples

draco/asp/hard.lp
Zehua-Zeng: @domoritz When I translated from Draco1, I was also not sure about these two.

draco/asp/hard.lp
Zehua-Zeng: @domoritz

draco/asp/hard.lp
domoritz: I would also have to dig into the meaning. When in doubt, leave it out (and then see when we generate bad visualizations).

draco/asp/hard.lp
domoritz: Agreed. But too many colors is bad.

draco/tests/test_hard.py
Zehua-Zeng: I trimmed the examples to make them shorter

draco/asp/hard.lp
Zehua-Zeng: I removed them for now

draco/asp/hard.lp
Zehua-Zeng: I think we should still keep these two to avoid ineffective recommendations

domoritz: Could be a soft constraint if we are unsure.

Zehua-Zeng: removed

PR #269, opened 2022-04-11
docs: add the guide for how to add new constraints

JunranY:


docs/guide.ipynb
```diff
@@ -0,0 +1,180 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "mechanical-pointer",
+   "metadata": {},
+   "source": [
+    "# How to customize or contribute to the knowledge base\n",
+    "\n",
+    "The [asp](https://github.com/cmudig/draco2/tree/main/draco/asp) directory contains the Draco knowledge base as answer set programs. \n",
```
domoritz, line 10: I think this should go into https://github.com/cmudig/draco2/blob/main/CONTRIBUTING.md. I imagine in the docs we should talk about how someone could add custom constraints assuming they install Draco via pypi (which we can't do yet since we haven't uploaded Draco yet). We could still link to the list of files in the asp directory from here since it's useful inspiration.

docs/guide.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    " * `soft.lp` defines soft constraints in the form of violation/1 and violation/2 predicates. By themselves, these predicates don't change the search.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "promising-steal",
+   "metadata": {},
+   "source": [
+    "## guideline\n",
+    "You can design your own description language and use it with Draco or extend the existing language we use here. If you don't know where to start with the constraints, you can first set up the search space in `generate.lp` and use our `run_clingo` API to generate some recommendations. Then, you should be able to find some recommendations that should have been left out, and you can write constraints to reflect them.  \n",
```
domoritz, line 27: Slight modify this so it doesn't assume that you can modify generate.lp.

docs/guide.ipynb
```diff
@@ -0,0 +1,180 @@
...
+    "from pprint import pprint\n",
+    "\n",
+    "prog = (\n",
+    "    generate.program\n",
+    "    + define.program\n",
+    "    + helpers.program\n",
+    "    + hard.program\n",
+    "    + constraints.program\n",
+    ")\n",
+    "with open(\"../draco/asp/examples/scatter.lp\") as file:\n",
```
domoritz, line 73: Let's define the program we want to demonstrate in the code instead of loading it from a file.

domoritz (changes requested):

docs/guide.ipynb
JunranY: Do you mean I should split this doc and put this part in contributing.md or just move this entire thing?

JunranY: moved the links to programs docs

docs/guide.ipynb
JunranY: I didn't mean they can modify the generate.lp from the programs.
I'm thinking about when they design their own language, adding say entity types we don't define and writing their new generator. Is this outside of what you imagined for draco2?

domoritz: Your right that that's in scope. We already have a place to document existing programs, though, so I want to make sure we don't repeat content.

docs/guide.ipynb
```diff
@@ -0,0 +1,191 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "mechanical-pointer",
+   "metadata": {},
+   "source": [
+    "# How to customize or contribute to the knowledge base"
```
domoritz, line 8:
```suggestion
    "# How to customize or extend the knowledge base"
```

docs/guide.ipynb
```diff
@@ -0,0 +1,191 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8866d38f",
+   "metadata": {},
+   "source": [
+    "If you see that there are too many recommendations, you can:\n",
+    " * add more hard constraints\n",
+    " * modify your generator and hard constraints to reduce symmetry in the search space (ex. similar recommendations with switched entity ids)\n",
```
domoritz, line 102:
```suggestion
    " * modify your generator and hard constraints to reduce symmetry in the search space (e.g. similar recommendations with switched entity ids)\n",
```

docs/guide.ipynb
```diff
@@ -0,0 +1,191 @@
...
+    "\n",
+    "\n",
+    "If you see too few recommendations, you can:\n",
+    " * check if some of your constraints are too tight, and move them to soft constraints\n",
+    " \n",
+    "\n",
+    "If you see no recommendations, you might have made mistakes in the hard constraints. You can allow violations to check what are the common ones by doing the following:"
```
domoritz, line 109: Maybe say what you can do and then point to the example. For example "You can allow violations to check what are the common ones by .... Below is an example:"

domoritz (approved):

PR #273, opened 2022-04-12
feat: soft constraints for scale type

Zehua-Zeng:


draco/asp/soft.lp
```diff
@@ -30,3 +30,18 @@ preference(same_field,F) :-
     entity(mark,_,M),
     { entity(encoding,M,E): attribute((encoding,field),E,F) } = 2.
 
+% @soft(type_linear) Prefer linear=log>ordinal>categorical
```
JunranY, line 33: I think in draco1 these are about data types and yours are scale types

JunranY (changes requested):

draco/asp/soft.lp
domoritz: Yes. We don't need endowing types anymore.

domoritz (approved):

JunranY (approved):

PR #275, opened 2022-04-13
feat: some soft constraints in 

Zehua-Zeng:


domoritz (changes requested):

draco/tests/test_soft.py
```diff
@@ -332,8 +329,607 @@ def test_same_field():
...
+
+    entity(scale,root,s1).
+    attribute((scale,channel),s1,x).
+    attribute((scale,type),s1,categorical).
+    """
+        )
+        == [("number_categorical", "temperature")]
+    )
+
+    # two scales, categorical for number
```
Zehua-Zeng, line 640: @domoritz

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+    entity(field,_,F),
+    entity(mark,_,M),
+    { entity(encoding,M,E): attribute((encoding,field),E,F) } >= 3.
+
+% @soft(count_grt1) Prefer not to use count more than once.
+preference(count_grt1,M) :-
+    entity(mark,_,M),
+    { entity(encoding,M,E): attribute((encoding,aggregate),E,count) } > 1.
+
+% @soft(shape_high_cardinality) Shape channel should not have too high cardinality.
```
JunranY, line 44: I think you need to include that the scale is discrete(discrete data or binned). 
the `mark_channel_discrete_or_binned` helper might be relevant.

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+    N > 5.
+
+% @soft(number_categorical) Numbers should not be nominal.
+preference(number_categorical,F) :-
+    attribute((field,type),F,number),
+    attribute(field_scale,F,categorical).
+
+% @soft(bin_number_low_cardinality) Binned number field should not have too low cardinality.
+preference(bin_number_low_cardinality,E) :-
+    attribute((field,type),F,number),
```
JunranY, line 57: datetime might also work as quantitative data type
```suggestion
    attribute((field,type),F,(number;datetime)),
```

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+    attribute((encoding,field),E,F),
+    attribute((encoding,binning),E,B),
+    B < 15.
```
JunranY, line 60: also I think this constraint should be about data cardinality(unique) rather than the number of bins

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+
+% @soft(bin_not_linear) Prefer linear scale for bin.
+preference(bin_not_linear,E) :-
```
JunranY, line 63: How about bin-ordinal for the color channel

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+    attribute((encoding,binning),E,_),
+    not attribute((encoding,scale_type),E,linear).
+
+% @soft(only_discrete) Plots with only nominal, ordinal, binned q, or t with time unit should add either an aggregation (e.g. count) or a quantitative field.
```
JunranY, line 67: The constraint doesn't match the description here. I think the description says that every encoding of the plot is discrete, but here you are only saying the encodings for one mark are discrete.

draco/asp/soft.lp
```diff
@@ -30,18 +30,67 @@ preference(same_field,F) :-
...
+preference(only_discrete,M) :-
+    attribute(mark_channel_discrete_or_binned,M,_).
+
+% @soft(multi_non_pos) Prefer not to use multiple non-positional encoding channels.
+preference(multi_non_pos,M) :-
```
JunranY, line 72: I'm not sure but I think we need another constraint about non-positional channels in views since this one is about marks.

draco/asp/soft.lp
Zehua-Zeng: Isn't it even number data type (continuous) can also have low cardinality, it doesn't necessarily to be discrete.

draco/asp/soft.lp
Zehua-Zeng: There is a hard constraint that binning can only be added to number/datatime. And also there are two soft constraints bin_low and bin_high, so I wonder if I should keep this one instead.

draco/asp/soft.lp
Zehua-Zeng: I guess bin-linear is the most preferred combination, so if not linear there is a penalty. Or we can split into bin-linear, bin-ordinal, bin-log, bin-categorical, and then assign different weights.

draco/asp/soft.lp
Zehua-Zeng: changed discription

draco/asp/soft.lp
Zehua-Zeng: I think using multi non-pos channels in one mark already has a penalty, then a view with multiple marks having this situation will have multiple penalties?

draco/asp/soft.lp
JunranY: but scale for shape need to be discrete? maybe we can let a hard constraint handle this

draco/asp/soft.lp
JunranY: but soft constraint is doing a different thing from hard constraint. I think you still need to include the datatime for the purpose of counting the preference.

draco/asp/soft.lp
```diff
@@ -64,7 +64,7 @@ preference(bin_not_linear,E) :-
     attribute((encoding,binning),E,_),
     not attribute((encoding,scale_type),E,linear).
 
-% @soft(only_discrete) Plots with only nominal, ordinal, binned q, or t with time unit should add either an aggregation (e.g. count) or a quantitative field.
+% @soft(only_discrete) Only discrete encoding channels are used in a mark.
```
JunranY, line 67: I think the previous description is also a valid soft constraint though

draco/asp/soft.lp
JunranY: I mean for example you can have one non-positional for layer 1 and a different non-positional for layer 2, which doesn't violate this particular preference. but for such view, it has 2 non-positional in total.

draco/asp/soft.lp
Zehua-Zeng: I typed a word wrongly lol. I meant maybe we should *remove* this soft constraint since there are hard and other similar soft

draco/asp/soft.lp
Zehua-Zeng: I am not sure whether two non-pos in two separate layers should be penalized or not...

draco/asp/soft.lp
JunranY: but `bin_low/high` is about the number of bins; this one is about field cardinality

JunranY (changes requested):

domoritz:
Please make the issue title more descriptive of the content.

draco/tests/test_soft.py
domoritz: What can I help with?

draco/asp/helpers.lp
```diff
@@ -59,3 +59,16 @@ attribute(mark_channel_discrete_or_binned,M,C) :-
...
+% @helper(encoding_cardinality,E,N) The cardinality of an encoding.
+attribute(encoding_cardinality,E,N) :-
+    attribute((field,unique),F,N),
+    attribute((encoding,field),E,F),
+    not attribute((encoding,binning),E,_).
+attribute(encoding_cardinality,E,N) :-
+    attribute((encoding,binning),E,N).
+
+% @helper(field_scale,F,T) The scale type for a field.
+attribute(field_scale,F,T): :-
```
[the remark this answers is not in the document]
domoritz:
```suggestion
attribute(field_scale,F,T) :-
```

domoritz:
Regarding the title, what is an encoding preference?

Zehua-Zeng:
> Regarding the title, what is an encoding preference?

Like shape_high_cardinality, multiple non-position encodings, use non-pos before position?

domoritz:
Ahh ok. Sounds good to me.

draco/asp/hard.lp
```diff
@@ -53,6 +53,11 @@ violation(count_without_q) :-
     attribute((encoding,scale_type),E,T),
     domain(discrete_scale,T).
 
+% @hard(shape_not_ordinal) Shape requires discrete and ordinal (ordinal/categorical doesnt matter, so we use oridnal only here).
```
JunranY, line 56:
```suggestion
% @hard(shape_not_ordinal) Shape requires discrete and ordinal (ordinal/categorical doesn't matter as scale types, so we use oridnal only here).
```

JunranY (approved):

draco/asp/helpers.lp
```diff
@@ -59,3 +59,16 @@ attribute(mark_channel_discrete_or_binned,M,C) :-
...
+
+% @helper(encoding_cardinality,E,N) The cardinality of an encoding.
+attribute(encoding_cardinality,E,N) :-
+    attribute((field,unique),F,N),
+    attribute((encoding,field),E,F),
+    not attribute((encoding,binning),E,_).
+attribute(encoding_cardinality,E,N) :-
+    attribute((encoding,binning),E,N).
+
+% @helper(field_scale,F,M,T) The scale type for a field.
```
domoritz, line 71: There is no 1:1 mapping from scales to fields so I don't think we can use this.

draco/asp/soft.lp
```diff
@@ -30,18 +30,68 @@ preference(same_field,F) :-
...
+    { entity(encoding,M,E): attribute((encoding,aggregate),E,count) } > 1.
+
+% @soft(shape_high_cardinality) Shape channel should not have too high cardinality.
+preference(shape_high_cardinality,E) :-
+    attribute((encoding,channel),E,shape),
+    attribute(encoding_cardinality,E,N),
+    N > 5.
+
+% @soft(number_categorical) Numbers should not be nominal.
+preference(number_categorical,F,M) :-
```
domoritz, line 51: I think if you add the encoding to this constraint, then you don't need the helper.

draco/tests/test_soft.py
```diff
@@ -332,8 +329,681 @@ def test_same_field():
     )
 
 
-def test_type_linear():
-    b = soft.blocks["type_linear"]
+def test_same_field_grt3():
```
domoritz, line 332: Do you think these tests are good at actually finding potential issues? I think we need to write the test cases in a way where they can reasonably fail. I'll add comments below.

draco/tests/test_soft.py
```diff
@@ -332,8 +329,681 @@ def test_same_field():
...
+
+    attribute((encoding,field),e1,temperature).
+    attribute((encoding,field),e2,temperature).
+    attribute((encoding,field),e3,temperature).
+    """
+        )
+        == [("same_field_grt3", "temperature")]
+    )
+
+    assert (
```
domoritz, line 390: This case is almost the same as above so we could remove it.

draco/tests/test_soft.py
```diff
@@ -332,8 +329,681 @@ def test_same_field():
...
+    attribute((encoding,field),e5,date).
+    attribute((encoding,field),e6,date).
+    """
+        )
+        == [("same_field_grt3", "temperature"), ("same_field_grt3", "date")]
+    )
+
+    # use field temperature three times with mark m1,
+    # and field date with 2 different marks
+    assert (
```
domoritz, line 419: To simplify, we could remove temperature and only concentrate on the case with date and two marks. Simplifies the tests a bit.

draco/tests/test_soft.py
domoritz: The main reason I suggest it is because the file is getting pretty long and I worry about maintainability. 

I'm open to ideas.

domoritz (changes requested):

domoritz (approved):

Issue #274, opened 2022-04-12
Add a utility method to get the constants of a clingo program as a dictionary


domoritz:

JunranY:
#280

PR #280, opened 2022-04-19
feat: add a util method to get the constants dict(#274), port the weights and others to support ranking.

JunranY:
I copied the process_weight form draco1, but how about we parse the weights.lp, process it and generate the assign_weights functions so the users don't have to?


JunranY:
The coverage is less than 100% because of `process_soft.lp`. 
I'm not sure how to test it when it reads and writes to specific files.

domoritz:
> I copied the process_weight form draco1, but how about we parse the weights.lp, process it and generate the assign_weights functions so the users don't have to?

I think that makes sense. It would make Draco a lot easier to extend as well. 

> Also not sure if we want to keep it or change the format to make it more integrated with the rest of the draco2 module

Yes, that would be great.

draco/programs.py
```diff
@@ -31,3 +31,6 @@ def get_program(file_path: Path) -> Program:
 hard = get_program(asp_path / "hard.lp")
 helpers = get_program(asp_path / "helpers.lp")
 soft = get_program(asp_path / "soft.lp")
+assign_weights = get_program(asp_path / "assign_weights.lp")
```
domoritz, line 34: Instead of generating the file, can we instead only do this in memory?

draco/asp_utils.py
```diff
@@ -91,3 +91,19 @@ def blocks_to_program(
...
     ]
+
+
+def get_constants(program: Union[str, Iterable[str]]) -> Dict:
+    if isinstance(program, str):
+        program = program.split(".")
+
+    constants = {}
+    for line in program:
+        match = re.search("#const (.*) = ([-0-9]*)", line)
```
domoritz, line 102: Instead of running a regex search, can we use the lingo AST or the lingo API?

draco/process_weights.py
```diff
@@ -0,0 +1,76 @@
...
+            if match:
+                assign_prog += f"preference_weight({match.group(1)},{name}).\n"
+
+    with open(assign_path, "w") as assign:
+        assign.write(assign_prog)
+
+    return WeightsAssign(const_prog, assign_prog, const_dict)
+
+
+def set_new_weights(
```
domoritz, line 44: I don't think we should have this method. The problem with writing the weights file is that if someone does this when they import the library, the next time they use the library, the weights will be different. This breaks a fundamental assumption that libraries are stateless.

domoritz (changes requested):

draco/programs.py
JunranY: removed this line.
`assign_weights` now is an instance variable of `WeightsAssign`

draco/asp_utils.py
JunranY: updated the code using the AST API. 
I'm not sure if this is bad because I'm not familiar with the clingo AST elements and they don't have a lot documentation. So, thanks for the review!

draco/process_weights.py
JunranY: removed it. The reason I added it was that we already made it possible to change the constant values temporarily in `run_clingo`. Besides, the users are also able to directly modify the .lp files including weights.lp, so I thought we could provide a method for that.

domoritz (approved):

PR #283, opened 2022-04-26
feat: soft constraints for aggregate and zero

Zehua-Zeng:
Also added sorting the list_preference output


draco/asp/helpers.lp
```diff
@@ -60,10 +67,25 @@ attribute(mark_channel_discrete_or_binned,M,C) :-
...
 % @helper(encoding_cardinality,E,N) The cardinality of an encoding.
 attribute(encoding_cardinality,E,N) :-
     attribute((field,unique),F,N),
     attribute((encoding,field),E,F),
     not attribute((encoding,binning),E,_).
 attribute(encoding_cardinality,E,N) :-
     attribute((encoding,binning),E,N).
+
+% @helper(mark_has_aggregate) Whether a mark has aggregate encoding.
+attribute(mark_has_aggregate,M) :-
```
domoritz, line 89: I don't think we need this helper as we can just check `attribute((encoding,aggregate),E,_).`. Let's remove it. In general, we should only add helpers if it makes a big difference as every helper makes it harder to read the code.

domoritz:
@JunranY can you review this?

draco/asp/soft.lp
```diff
@@ -81,6 +81,69 @@ preference(non_pos_used_before_pos,M) :-
...
+
+% @soft(aggregate_no_discrete) Aggregate should also have a discrete encoding to group by.
+preference(aggregate_no_discrete,M) :-
+    entity(encoding,M,EA),
+    attribute((encoding,aggregate),EA,_),
+    not attribute(mark_encoding_discrete_or_binned,M,_).
+
+% @soft(x_y_raw) Prefer not to use plot with x and y discrete and no aggregate as it leads to occlusion.
+preference(x_y_raw,M) :-
+    attribute(mark_channel_discrete_or_binned,M,(x;y)),
```
JunranY, line 100: you are saying "either" with the `;`, but it should be "both"

JunranY (approved):

draco/asp/soft.lp
Zehua-Zeng: done

draco/tests/test_soft.py
```diff
@@ -7,11 +7,16 @@ def list_preferences(program: str):
...
+        pref_list = [
             tuple(map(lambda x: x.name, symbol.arguments))
             for symbol in model.answer_set
             if symbol.name == "preference"
         ]
+
+        pref_list.sort(key=lambda x: x[1])
+
+        return pref_list
+
```
domoritz, line 19: Please return `.sorted()` instead

domoritz (changes requested):

draco/asp/helpers.lp
```diff
@@ -60,10 +67,20 @@ attribute(mark_channel_discrete_or_binned,M,C) :-
...
+    attribute((encoding,binning),E,_).
+
 % @helper(encoding_cardinality,E,N) The cardinality of an encoding.
 attribute(encoding_cardinality,E,N) :-
     attribute((field,unique),F,N),
     attribute((encoding,field),E,F),
     not attribute((encoding,binning),E,_).
 attribute(encoding_cardinality,E,N) :-
-    attribute((encoding,binning),E,N).
+    attribute((encoding,binning),E,N).
```
domoritz, line 86: We usually have empty lines at the end of files.

draco/tests/test_soft.py
Zehua-Zeng: done

PR #290, opened 2022-05-05
feat: add field entropy and soft constraint weights

Zehua-Zeng:


draco/asp/weights.lp
```diff
@@ -7,7 +7,43 @@
...
+#const date_not_x_weight = 1.
+#const x_row_weight = 1.
+#const y_row_weight = 1.
+#const x_col_weight = 1.
+#const y_col_weight = 1.
+
+#const linear_scale_weight = 0.
+#const log_scale_weight = 1.
+#const ordinal_scale_weight = 2.
+#const categorical_scale_weight = 3.
```
JunranY, line 49:
```suggestion
#const categorical_scale_weight = 3.

```

JunranY (approved):

docs/facts/schema.md
```diff
@@ -55,10 +59,12 @@ attribute((field,unique),f2,111).
 attribute((field,min),f2,0).
 attribute((field,max),f2,55).
 attribute((field,std),f2,6).
+attribute((field,entropy),f1,2.4218).
```
domoritz, line 62: Hmm, this doesn't parse for me. 

Screen Shot 2022-05-06 at 7 15 58 PM

domoritz (changes requested):

docs/facts/schema.md
Zehua-Zeng: Yeah, I realized that when I wrote soft constraints with entropy. I wonder how I can fix this issue? Should I change the entropy to int instead?

JunranY: I think you can use the string format

domoritz: String format won't let us do comparisons. I think we need to use int.

PR #291, opened 2022-05-08
feat: soft constraints for Kim entropy preferences and stacking

Zehua-Zeng:
In original Draco 1 soft constraints, high entropy is defined as EN > 12, however, I barely see entropy larger than 5. So I change it to 5. Open to discussion.


JunranY (approved):

draco/asp/soft.lp
```diff
@@ -261,6 +261,38 @@ preference(y_col,V) :-
     entity(facet,V,F),
     attribute((facet,channel),F,col).
 
+% @soft(color_entropy_high) Entropy, primary quantitaty interactions as suggested by Kim et al.
+preference(color_entropy_high,E) :-
+    attribute((field,entropy),F,EN),
+    attribute((encoding,field),E,F),
+    attribute((encoding,channel),E,color),
+    attribute((encoding,scale_type),E,linear),
```
JunranY, line 269: maybe this instead:
```suggestion
    attribute((field,type),E,number;datetime),
```

PR #298, opened 2022-05-13
feat: soft constraints for encoding and scale correlations

Zehua-Zeng:


JunranY (commented):
I'm not sure if these are necessary. I think assigning weights is difficult  because they might be case-by-case for different designs?

Zehua-Zeng:
> I'm not sure if these are necessary. I think assigning weights is difficult because they might be case-by-case for different designs?

@domoritz

domoritz:
These are also in draco 1? I can see them being useful to tune a model. Yes, there may not be a universally good value for all of these but it provides more features to tune. We could set the default to 0 for most of them.

PR #309, opened 2022-05-31
fix: generate `(scale,zero)`; fix `only discrete`

JunranY:


draco/asp/examples/histogram.lp
```diff
@@ -3,10 +3,15 @@ attribute(number_rows,root,100).
...
+attribute((field,unique),(f,0),100).
 
-entity(mark,root,(m,0)).
+entity(view,root,(v,0)).
+
+entity(mark,(v,0),(m,0)).
 entity(encoding,(m,0),(e,0)).
 attribute((encoding,field),(e,0),(f,0)).
 
+entity(encoding,(m,0),(e,1)).
```
domoritz, line 14: I think this should be auto generated.

domoritz (changes requested):

domoritz (approved):

PR #312, opened 2022-06-06
feat: add overlap 

JunranY:


draco/asp/helpers.lp
```diff
@@ -106,24 +116,96 @@ attribute(is_c_c,M) :-
...
+    attribute(mark_encoding_cont,M,E),
+    attribute((encoding,channel),E,y),
+    not attribute((mark,channel),M,x).
+attribute(discrete_size,M,N) :-
+    attribute(is_d_d,M),
+    attribute(x_y_cardinality,M,x,NX),
+    attribute(x_y_cardinality,M,y,NY),
+    N = NX*NY.
+
+% @helper(x_y_cardinality) to go from quadratic to linear number of grounding
```
domoritz, line 203:
```suggestion
% @helper(x_y_cardinality) Cardinality of x and y. Helps to go from quadratic to linear number of grounding.
```

draco/asp/helpers.lp
```diff
@@ -106,24 +116,96 @@ attribute(is_c_c,M) :-
...
+attribute(x_y_cardinality,M,x,N) :-
```
domoritz, line 204: Since x and y is not consistent here, should we have `attribute(x_cardinality,M,N)` or `attribute(cardinality,M,x,N)`?

domoritz (approved):

draco/asp/soft.lp
```diff
@@ -334,33 +334,80 @@ preference(c_c_text,M) :-
 % @soft(c_d_point) Continuous by discrete for point mark.
 preference(c_d_point,M) :-
     attribute(is_c_d,M),
+    not attribute(no_overlap,M),
```
domoritz, line 337: it should not be the case that there is not no overlap = it should not be the case that there is overlap. Can we use a different helper to make this simpler?

domoritz: What's the point of adding the overlap fact into this preference?

draco/asp/examples/histogram.lp
```diff
@@ -7,9 +7,10 @@ attribute((field,unique),(f,0),100).
 
 entity(view,root,(v,0)).
 
-entity(mark,(v,0),(m,0)).
-entity(encoding,(m,0),(e,0)).
-attribute((encoding,field),(e,0),(f,0)).
+entity(mark,(v,0),(m,0,0)).
+entity(encoding,(m,0,0),(e,0,0,0)).
+attribute((encoding,field),(e,0,0,0),(f,0)).
+attribute((encoding,binning),(e,0,0,0),10).
```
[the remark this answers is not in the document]
JunranY: without specifying binning, it gives a 1d tick plot. Adding binning makes it weighing among "number of encodings", "aggregation", "tick or bar", and "overlapping or not".

draco/asp/soft.lp
JunranY: I think `c_d_point` and `c_d_point_no_overlap` need to be mutually excluded. Otherwise `c_d_point_no_overlap` will be penalized twice for both `c_d_point` and `c_d_point_no_overlap`?

domoritz: It doesn't have to be. If we have two weights for mutually exclusive preferences an and b, then we can always simulate that with with a negative weight. It may make sense to make constraints mutually exclusive and we should re-evaluate as we improve the weights.

draco/asp/generate.lp
```diff
@@ -8,21 +8,22 @@
 % single view v0 on the root
 { entity(view,root,(v,0)) } = 1.
 
-% maximum number of non-layered views.
+% % maximum number of non-layered views.
```
domoritz, line 11:
```suggestion
% maximum number of non-layered views.
```

draco/asp/examples/histogram.lp
```diff
@@ -10,6 +10,7 @@ entity(view,root,(v,0)).
 entity(mark,(v,0),(m,0)).
 entity(encoding,(m,0),(e,0)).
 attribute((encoding,field),(e,0),(f,0)).
+attribute((encoding,binning),(e,0),10).
```
domoritz, line 13: Why do we need to specify the binning here>? Shouldn't it be automatically be added by generate?

draco/asp/examples/histogram.lp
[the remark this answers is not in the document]
domoritz: Binning should be added by generate (not required but optional). That's how Draco 1 worked as well.

PR #327, opened 2022-07-01
fix: downgrade pandas for pandas-stubs

JunranY:


pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 clingo = "^5.5.2"
-pandas = "^1.4.2"
+pandas = "1.4.2"
```
domoritz, line 30: This is fixing the pandas version and not the stubs. I don't think we want to be so restrictive if we can avoid it.

JunranY: I did it because of this 
image
what should I do instead?

domoritz: I updated the lock file in https://github.com/cmudig/draco2/pull/330. Let's see whether it passes.

Issue #332, opened 2022-07-04
Scatterplot recommendation generates four scales


domoritz:
In the recommendation for a scatterplot we get four scales. We should not have the ordinal scales here. 

See https://github.com/cmudig/draco2/pull/331/files#diff-fd529544f65ea5ca3eaa03f3fd5db949f51be3dc4e3b937aba79198e3c1dcea1 for how to reproduce this issue.

```
OUTPUT:
{'field': [{'name': 'temperature', 'type': 'number'},
           {'name': 'precipitation', 'type': 'number'}],
 'number_rows': 100,
 'task': 'summary',
 'view': [{'mark': [{'channel': 'y',
                     'encoding': [{'channel': 'x',
                                   'field': ('f', 0),
                                   'scale_type': 'linear',
                                   'type': 'linear',
                                   'zero': 'true'},
                                  {'channel': 'y',
                                   'field': ('f', 1),
                                   'scale_type': 'linear',
                                   'type': 'linear',
                                   'zero': 'true'}],
                     'mark_channel_cont': 'y',
                     'mark_encoding_cont': ('e', 1),
                     'scale': ('e', 0),
                     'type': 'point'}],
           'scale': [{'type': 'ordinal'},
                     {'type': 'ordinal'},
                     {'channel': 'y',
                      'field': ('f', 1),
                      'scale_type': 'linear',
                      'type': 'linear',
                      'zero': 'true'},
                     {'channel': 'x',
                      'field': ('f', 0),
                      'scale_type': 'linear',
                      'type': 'linear',
                      'zero': 'true'}]}]}
```

PR #334, opened 2022-07-06
fix: scale generation for scatterplot

JunranY:
fixed issue #332 

One potential problem is that the partial spec need to follow the format `(s, 0)` for scale ids. 
For other entities like marks or encoding, following the format (eg. `(m,0)`) can guarantee the max number of such entity. Otherwise, it would refer to the max number of **additional** entity. 
I'll try to fix this.


domoritz:
It's fine if we use the special format for additional entities. We can always limit the actual number of entities with an aggregate constraint.

draco/asp/generate.lp
```diff
@@ -38,10 +36,14 @@ not_required((encoding,aggregate)).
 not_required((encoding,field)).
 not_required((encoding,stack)).
 
-% generator(scale) generate scales such that their ids and channels corresponde to the encodings in each view.
-{ entity(scale,V,E) } = 1 :- entity(mark,V,M), entity(encoding,M,E), attribute((encoding,channel),E,C).
-attribute((scale,channel),S,C) :- entity(scale,V,S), attribute((encoding,channel),S,C).
+% @generator(scale) generate scales for all encoding channels from each view.
+% count the total number of channels of the marks in the same view.
+count_channels(V,N) :- N = #count{ C : domain(encoding_channel,V,C) }, entity(view,_,V).
```
domoritz, line 41: We should try to avoid putting the count into a fact. The problem with doing that is that the grounding can get large. Could we reuse the encoding id somehow in the scale id generation? Id not, that's okay.

domoritz (approved):

draco/asp/generate.lp
JunranY: I tried using the encoding ids before which resulted the issue #332: some scale entities are included in the partial spec but ignored by the generator; it instead generates all new scales based the encodings. Do we always assume scale entities are not in the spec? If so, we are safe to use the encoding ids.

domoritz: What do you mean by "Do we always assume scale entities are not in the spec"?

domoritz: I think they can but I would not expect them to contain scales in most cases.

JunranY:
reuse encoding ids for scale id generation. 
hard constraints are used to ensure no duplicate channels. 
tested with partial specs that both include and exclude scales.

draco/asp/generate.lp
JunranY: @domoritz 
I just saw that the grounding was much better using the count than using encoding ids.
histogram: 5184 vs 7496
scatter: 8938 vs 12120

Should I revert it back?

PR #338, opened 2022-07-14
refactor: only assign the weights in a program instead of via args

domoritz:


domoritz:
Ah, we actually need it. The assign weights only assigns the weights to the soft constraints but does not set the weight.

PR #340, opened 2022-07-19
docs: improve docs

peter-gy:
### Goals

- fix various typos in the docs and source files
- add hyperlinks in docs where sensible
- make the docs overall more pleasant to read

### Notes

The [documentation](https://dig.cmu.edu/draco2/api/draco.html) of the `draco.Draco` class is pretty verbose thanks to the automatic default value expansion of Sphinx. I think it would make sense to spawn a new issue for its fix (I would be glad to contribute).  I have two approaches on my mind for the fix:

1. Tap into the Jupyter Book configuration and register a Sphinx function to remove the default value from the expansion, as suggested by [this StackOverflow answer](https://stackoverflow.com/a/69865409/14457639)
2. Create a custom `__repr__` and/or `__str__` in `programs.Program` and `weights.Weights` to have a more compact value representation, returning a summary of the object instead of the stringified fields generated by `dataclass`

What do you think?

### DIFF
--- a/docs/api/draco.ipynb
+++ b/docs/api/draco.ipynb
@@ -13,7 +13,11 @@
   {
    "cell_type": "markdown",
    "id": "c61fd747",
-   "metadata": {},
+   "metadata": {
+    "pycharm": {
+     "name": "#%% md\n"
+    }
+   },
    "source": [
     "## Class Documentation\n",
     "\n",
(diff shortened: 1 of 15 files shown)


domoritz:
Thank you! And yes, I noticed that the docs for `draco.Draco` rendered very poorly. I tried fixing it the other day but didn't get anywhere. I'd very much welcome a pull request.

PR #343, opened 2022-07-20
docs: no default value expansion

peter-gy:
### Goals

- Disable the expansion of default values in docstrings
- Fix the looks of the [Draco Class Documentation section](https://dig.cmu.edu/draco2/api/draco.html#class-documentation)

_Users_petergy_Coding_OpenSourceProjects_draco2_docs__build_html_api_draco html

### Notes

 - Found a solution which is the cleanest one imho: there is an [official `autodoc` config](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_preserve_defaults) for this

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,27 +12,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up Python
-        uses: actions/setup-python@v4
       - name: Set up Node
         uses: actions/setup-node@v3
-      - name: Install Poetry
-        uses: snok/install-poetry@v1.3
-      - name: Cache Poetry virtualenv
-        uses: actions/cache@v3.0.4
-        id: cache
+      - name: Install poetry
+        run: pipx install poetry
+      - uses: actions/setup-python@v4
         with:
-          path: ~/.virtualenvs
-          key: poetry-${{ hashFiles('**/poetry.lock') }}
-          restore-keys: |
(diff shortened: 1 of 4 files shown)


peter-gy:
The GitHub workflow fails at [step 6](https://github.com/cmudig/draco2/runs/7425525058?check_suite_focus=true#step:9:7), however, the tests succeed for me locally.

```
Run poetry run pytest --cov=./ --cov-report=xml draco
Creating virtualenv draco-2B3FyYM-py3.8 in /home/runner/.virtualenvs
```

I assume that the failure is caused by the fact that no cached virtualenv is found by the time `poetry run pytest --cov=./ --cov-report=xml draco` gets executed, hence, `pytest` is not available for running the tests.

However, this problem is most probably not related to the changes I introduced in this PR.

domoritz:
I tried restarting the builds and they still fail. I haven't seen this before.

peter-gy:
The GitHub action fix I proposed handles the installation of `poetry` and the package caching via `actions/setup-python@v4`, as described in [its docs](https://github.com/actions/setup-python#caching-packages-dependencies). Before considering accepting this change, please run the workflow manually once again to make sure that the caching works as expected. (It works in my fork)

domoritz:
Hmm, now I don't see any of the methods documented at https://dig.cmu.edu/draco2/api/draco.html. Do you know how we can add it back?

peter-gy:
I actually don't recall that any of the methods have been documented in the previous version(s), but sure, I can take a look at it.

PR #320, opened 2022-06-21
feat: define.lp and example for pie charts

Zehua-Zeng:

PR #352, opened 2022-07-26
feat: add coordinate systems

Zehua-Zeng:


draco/asp/define.lp
```diff
@@ -3,6 +3,9 @@
 % @definition(mark_type) Types of marks to encode data.
 domain((mark,type),(point;bar;line;area;text;tick;rect)).
 
+% @definition(coordinate_type) Types of coordinates.
+domain((coordinate,type),(cartesian;polar)).
```
domoritz, line 7: Shouldn't this be

```suggestion
domain((view,coordinate_system),(cartesian;polar)).
```

domoritz:
Looking at these two options (this one and #320) , I feel this approach here with adding a coordinate system is simpler and cleaner. What do you think? If you don't have objections, I think we should merge this one.

Zehua-Zeng:
> @JunranY @Zehua-Zeng Looking at these two options (this one and #320) , I feel this approach here with adding a coordinate system is simpler and cleaner. What do you think? If you don't have objections, I think we should merge this one.

I agree. I already fixed the conflicts, and now it is ready to merge.

domoritz (approved):

docs/facts/examples.ipynb
```diff
@@ -118,6 +119,7 @@
     "    {\n",
     "        \"view\": [\n",
     "            {\n",
+    "                \"coordinate\": \"cartesian\",\n",
```
domoritz, line 122: This should not be required to be set manually, no?

domoritz (approved):

docs/facts/examples.ipynb
Zehua-Zeng: This is not required to be set manually. The recommender can pick based on soft constraints. I just wanna add to examples to clarify which coordinate system the chart is using.

domoritz: I would prefer to remove unnecessary details.

docs/facts/view.md
```diff
@@ -6,14 +6,16 @@ If a view has multiple marks, Draco assumes that the marks are layered (i.e. the
 
 ## View Properties
 
-By default, views don't have any properties.
+`(view,coordinate)`
```
domoritz, line 9:
```suggestion
`(view,coordinates)`
```

or

```suggestion
`(view,coordinate_system)`
```

domoritz (approved):
Thank you. Looks great with one naming change.

docs/facts/view.md
Zehua-Zeng: Done

PR #377, opened 2022-08-23
docs: add debug draco page

Zehua-Zeng:
count soft constraints violations for example charts


domoritz (commented):
This is great!

* Add a row for the weights (or an aligned chart)
* Use Altair for the chart
* Provide different sorting options (by sum, sum of non-zero, alphabetic, etc)

docs/applications/data/example_charts.json
```diff
@@ -0,0 +1,464 @@
+{
+  "tick_plot": [
```
domoritz, line 2: Maybe this is easier/more maintainable with [TOML](https://toml.io/en/)/YAML

domoritz (changes requested):
See comments

docs/applications/debug_draco.ipynb
```diff
@@ -0,0 +1,138 @@
...
+     "metadata": {
+      "needs_background": "light"
+     },
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "fig, ax = plt.subplots()\n",
+    "\n",
+    "ax.matshow(feature_matrix, cmap=plt.cm.Blues)\n",
```
domoritz, line 84: Maybe try Altair

domoritz (changes requested):

docs/applications/data/example_charts.toml
```diff
@@ -0,0 +1,462 @@
+tick_plot = [
```
domoritz, line 1: Is there a way to write the spec as a single string?

domoritz (commented):
Added a comment. Would Altair work better than mpl in this case?

docs/applications/data/example_charts.toml
peter-gy: Sure, then we'll just need to split the string by line. But if we go that "far", wouldn't it be even easier to maintain the example charts if we had a folder inside which each file represents a chart?

peter-gy:
> Added a comment. Would Altair work better than mpl in this case?

I think so, yes.

Regarding *"Provide different sorting options (by sum, sum of non-zero, alphabetic, etc)"*, what would make more sense: an `ipywidget`-based interactively steerable approach, or simply outputting the result of the different options below each other?

The former approach would be nice, but would be sensible only if the docs are opened with Binder or served locally, the latter would be beneficial to display results statically via JupyterBook.

peter-gy:
What do you think about plotting weights & violation counts this way? Are there other sorting combinations that might be interesting?

visualization

domoritz:
> Regarding "Provide different sorting options (by sum, sum of non-zero, alphabetic, etc)", what do you think, what would make more sense: an ipywidget-based interactively steerable approach, or simply outputting the result of the different options below each other?

I was thinking a widget. But I'm not sure how that would render in the book. Need to try and decide.

domoritz:
> What do you think about plotting weights & violation counts this way? Are there other sorting combinations that might be interesting?

Beautiful. Two requests. First, make the upper chart have a smaller height. Second, can you use white for exactly 0? It would make it easier to interpret the chart.

Other than that, I think we can merge this.

docs/applications/data/example_charts.toml
domoritz: You already did what I meant. I meant each spec should be one string instead of an array. I'll try to be more precise next time I give feedback.

peter-gy:
> > What do you think about plotting weights & violation counts this way? Are there other sorting combinations that might be interesting?
> 
> Beautiful. Two requests. First, make the upper chart have a smaller height. Second, can you use white for exactly 0? It would make it easier to interpret the chart.
> 
> Other than that, I think we can merge this.

Implemented in a63de80

visualization

peter-gy:
> 



> > Regarding "Provide different sorting options (by sum, sum of non-zero, alphabetic, etc)", what do you think, what would make more sense: an ipywidget-based interactively steerable approach, or simply outputting the result of the different options below each other?
> 
> I was thinking a widget. But I'm not sure how that would render in the book. Need to try and decide.

I looked into it. While it works nicely when serving the notebook locally, the statically-built Jupyter Book site does not support interactivity by default, because we are attempting to call Python code. The Jupyter Book [docs](https://jupyterbook.org/en/stable/interactive/interactive.html?highlight=widget#ipywidgets) mention that for this use-case, remote kernels are necessary to interact with Python via the web site.

Screenshot 2022-09-10 at 19 21 53

domoritz:
> I looked into it. While it works nicely when serving the notebook locally, the statically-built Jupyter Book site does not support interactivity by default, because we are attempting to call Python code. The Jupyter Book [docs](https://jupyterbook.org/en/stable/interactive/interactive.html?highlight=widget#ipywidgets) mention that for this use-case remote kernels are necessary to interact with Python via the website.

I figured this would be an issue. Would we still have a default, though? I would be okay with having only a static default on the website and a dynamic version if someone runs the kernel.

peter-gy:
> > I looked into it. While it works nicely when serving the notebook locally, the statically-built Jupyter Book site does not support interactivity by default, because we are attempting to call Python code. The Jupyter Book [docs](https://jupyterbook.org/en/stable/interactive/interactive.html?highlight=widget#ipywidgets) mention that for this use-case remote kernels are necessary to interact with Python via the website.
> 
> I figured this would be an issue. Would we still have a default, though? I would be okay with having only a static default on the website and a dynamic version if someone runs the kernel.

I did not push the code for the widget, the notebook merged to `main` only has the static `VConcatChart` for now. Maybe as we progress with the project and extend the docs we can dedicate a PR to add more interactivity to the examples?

domoritz:
Yep, I saw that. We can always do follow up PRs even for smallest additions. Feel free to send one for the widget.

PR #379, opened 2022-08-24
feat: add interestingness and soft constraints

JunranY:


docs/applications/draco_learn.ipynb
```diff
@@ -362,13 +362,22 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
+<<<<<<< HEAD
```
domoritz, line 365: Merge conflict markers

domoritz (changes requested):

draco/asp/soft.lp
```diff
@@ -268,31 +279,35 @@ preference(color_entropy_high,E) :-
     attribute((encoding,field),E,F),
     attribute((encoding,channel),E,color),
     helper((encoding,scale_type),E,linear),
-    EN > 5000.
+    helper(enc_interesting,E),
+    EN > 12000.
```
domoritz, line 283: Why did you change these values?

draco/asp/soft.lp
```diff
@@ -626,145 +641,214 @@ preference(summary_rect,M) :-
...
+    helper(mark_encoding_discrete_or_binned,_,E),
+    helper(enc_interesting,E).
 
 % @soft(summary_discrete_text) Discrete text for summary tasks.
 preference(summary_discrete_text,E) :-
     attribute(task,root,summary),
     attribute((encoding,channel),E,text),
-    helper(mark_encoding_discrete_or_binned,_,E).
+    helper(mark_encoding_discrete_or_binned,_,E),
+    helper(enc_interesting,E).
```
domoritz, line 799: It feels strange to me to add interestingness to all these preferences. Won't that break if we don't have interestingness?

draco/tests/test_soft.py
```diff
@@ -5385,11 +5395,13 @@ def test_value_continuous_x():
         list_preferences(
             b.program
             + """
+    attribute((field,interesting),precipitation,true).
```
domoritz, line 5398: This seems wrong to me. I don't think we want to have people have to add interestedness to all fields just to make the recommendations work.

domoritz (changes requested):
I don't think we should require interestingness for all the fields. Also, please add docs.

draco/asp/soft.lp
JunranY: the updated one is the same as in draco1. The previous values was written by @Zehua-Zeng and the reason I remembered was that she observed in the kim2018 dataset, the entropy values never exceeded 12000? I changed it back to check the testing score for draco-learn.

draco/asp/soft.lp
JunranY: with the changes, these presences surely have different meaning now. However I also checked that the current version is what meant by draco1 ("for summary task, the interested column are preferred to be discrete text. I also changed the test cases for them to not be broken.

draco/tests/test_soft.py
JunranY: As I interpreted from the kim2018 dataset and their paper, the interestingness is associated with the tasks. something like "identifying the values for some field". So I think if people don't specify the task or interestingness, the recommendation would still work, but probably generate less precise results.

draco/asp/soft.lp
domoritz: Okay. I don't think the prior values were super thought out so your values may be better.

draco/asp/soft.lp
domoritz: I think we should have simpler constraints when possible. Would it hurt to remove the restriction to interesting fields from the constraints?

draco/tests/test_soft.py
domoritz: I see. Yes, I think it makes sense to say tasks need associated interesting fields.

JunranY: In addition there are soft constraints about only task and mark (without interestingness).

draco/asp/soft.lp
domoritz: I think we should have simpler constraints when possible. Do we use interestingness only with task constraints? Would it hurt to remove the restriction to interesting fields from constraints that don't involve tasks?

docs/facts/schema.md
```diff
@@ -39,6 +39,9 @@ the actual entropy).
 
 `(field,freq)` : The frequency of the most common value. Only used for strings.
 
+`(field,interesting)` : When the [`task`](task.md) regards specific fields, fields can be marked as relevant to the task
+with interesting.
```
domoritz, line 43:
```suggestion
`(field,interesting)` : When the [`task`](task.md) regards specific fields, fields can be marked as relevant to the task.
```

domoritz (approved):

PR #390, opened 2022-09-11
docs: interactive debugging

peter-gy:
Use `ipywidgets` to make the debugging chart sorting configuration selection interactive via a dropdown.

Follow-up of #377.


peter-gy:
Hey, @domoritz, I kept the static `VConcatChart` in the notebook to grant an instant overview of the output of all sorting configurations. The interactive section renders only the default selection on the web page.

Screenshot 2022-09-11 at 15 19 44

PR #391, opened 2022-09-11
refactor: type enhancements

peter-gy:
Adds explicit typing to functions in `schema.py`.

As I was looking through the codebase and the model classes to apify in preparation for a server PR, I realised that adding explicit typing where possible would help us generate more useful REST API docs.


draco/schema.py
```diff
@@ -46,25 +78,42 @@ def schema_from_dataframe(
...
 
 
-def schema_from_file(
-    file_path: Path, parse_data_type=dtype_to_field_type
-) -> Dict[str, Any]:
+def _construct_field_props(
+    name: str,
+    ty: FieldType,
+    unique: int,
+    entropy: float,
```
domoritz, line 93: Why pass in these two and compute the others?

draco/schema.py
```diff
@@ -1,15 +1,50 @@
...
+class BaseFieldProps(TypedDict):
+    """Properties shared by fields of all types in a `Schema`."""
+
+    name: str
+    type: FieldType
+    unique: int
+    entropy: float
+
+
+class NumberFieldProps(BaseFieldProps):
```
domoritz, line 21: I wouldn't consider this very stable API, though. We may add new properties from time to time. Just a note and no todo.

draco/schema.py
peter-gy: Given that these values were included in a column's properties regardless of its data type, I just kept their computation in `schema_from_dataframe`.  But since we have a `_construct_field_props` function, it is right that it makes more sense to handle all computations there.

I addressed this in 46358c9

draco/schema.py
peter-gy: I also assumed that we will extend the schema as we go, but this PR introduces a primitive typing base which we will be able to extend whenever adding new properties.

PR #395, opened 2022-09-16
refactor: debug notebook (#393)

### LINKED ISSUE #393: Add debug.py
Refactor the debugging notebook and extract the relevant logic and move it to the the Draco package itself. This makes it more reusable. 

Also, remove the redundant charts and only have the interactive one.

peter-gy:
### Goals

- Refactor debug-data-generating logic and plotting methods to `debug.py`
  - Created `DracoDebug` and `DracoDebugPlotter`
- Use the new debug API in the `debug_draco.ipynb` notebook
- Add [Binder](https://mybinder.org/) badge to `README.md` and to `debug_draco.ipynb` to make it easier to interact with the repo

Closes #393


README.md
```diff
@@ -15,22 +15,31 @@
...
 [![codecov](https://codecov.io/gh/cmudig/draco2/branch/main/graph/badge.svg)](https://codecov.io/gh/cmudig/draco2)
 [![Language grade: Python](https://img.shields.io/lgtm/grade/python/g/cmudig/draco2.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/cmudig/draco2/context:python)
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/cmudig/draco2/HEAD)
 
 **Work in Progress**
 
-Draco is a formal framework for representing design knowledge about effective visualization design as a collection of constraints. You can use Draco to find effective visualization visual designs or validate visualization designs. Draco's constraints are based on Answer Set Programming (ASP) and solved with the [Clingo](https://github.com/potassco/clingo) constraint solver. We also implemented a way to learn weights for the recommendation system directly from the results of graphical perception experiment. Draco v2 is a much improved version of the first iteration of [Draco](https://github.com/uwdata/draco).
+Draco is a formal framework for representing design knowledge about effective visualization design as a collection of
```
domoritz, line 22: Is this formatting something we can do automatically in e.g. a hook? Otherwise I suspect the formatting will get worse if someone makes an edit in the future.

README.md
```diff
@@ -15,22 +15,31 @@
 [![code style black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
 [![codecov](https://codecov.io/gh/cmudig/draco2/branch/main/graph/badge.svg)](https://codecov.io/gh/cmudig/draco2)
 [![Language grade: Python](https://img.shields.io/lgtm/grade/python/g/cmudig/draco2.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/cmudig/draco2/context:python)
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/cmudig/draco2/HEAD)
```
domoritz, line 18: Awesome.

draco/debug.py
```diff
@@ -0,0 +1,218 @@
...
+
+class DracoDebug:
+    __DEFAULT_DRACO__ = Draco()
+    __DEFAULT_FEATURE_NAMES__ = set(__DEFAULT_DRACO__.soft_constraint_names)
+    __DEFAULT_WEIGHTS__ = __DEFAULT_DRACO__.weights
+
+    def __init__(
+        self,
+        specs: dict[str, Iterable[str] | str],
+        draco: Draco | None = None,
```
domoritz, line 18: why don't we initialize Draco in the default argument here?

domoritz: then None would also not be needed anymore

draco/debug.py
```diff
@@ -0,0 +1,218 @@
...
+        :param weights: A ``dict`` associating weights with feature (constraint) names.
+                        Keys of this ``dict`` are expected to be of the form
+                        ``"{feature_name}_weight"``. If the supplied ``feature_names``
+                        set contains a feature name for which no entry is present in
+                        this ``dict``, the default weight of the feature will be used.
+        """
+        self.specs = specs
+        cls = DracoDebug
+        if feature_names is None:
+            feature_names = cls.__DEFAULT_FEATURE_NAMES__
```
domoritz, line 38: I think this could be a bug. We don't use the provided Draco instance so the features may be different.

draco/debug.py
```diff
@@ -0,0 +1,218 @@
...
+        return [cls.__pref_tuple_extended_with_weight(tpl, weights) for tpl in tuples]
+
+
+class ChartConfig(NamedTuple):
+    title: str
+    sort_x: alt.Sort | str
+    sort_y: alt.Sort | str
+
+
+class DracoDebugPlotter:
```
domoritz, line 147: Any idea how to test this? If not, we should consider leaving the code in the notebook.

README.md
peter-gy: I added a new entry to our pre-commit config & opened a separate PR for it (#396)

draco/debug.py
peter-gy: That's right, thank you for spotting. I addressed this in dabb05b by removing the option to enter custom weights and feature names in the `DracoDebug.__init__`, as it would just lead to confusing, inconsistent behaviour. Users will have the chance to configure the debugger just by dependency injecting a pre-configured `Draco` instance along with the specs.

draco/debug.py
peter-gy: I added tests in dabb05b checking for structural properties of the produced chart based on the input dataset and configs. What do you think about them?

draco/debug.py
peter-gy: I just wanted to be structurally consistent with how I handled default value settings for mutable args such as `feature_names`, but I realise that using a default arg for `draco` would make more sense.

PR #404, opened 2022-09-20
refactor: `python3.10` migration

peter-gy:
### Goals

- Mark `3.10` as the minimum version of the package
- Use new typing features
  -  Type Hinting Generics In Standard Collections such as `list[str]`, `set[int]` ([PEP 585](https://www.python.org/dev/peps/pep-0585))
  - Writing union types as `X | Y` ([PEP 604](https://www.python.org/dev/peps/pep-0604))
  - Explicit Type Aliases ([PEP 613](https://www.python.org/dev/peps/pep-0613))
- Handle dependabot dependency updates with conflict resolution 
  - #399 
  - #400 
  - #401


domoritz (approved):
Beautiful. Thank you.

peter-gy:
> Beautiful. Thank you.

No problem! I'll just merge this then and clean up the related dependabot PRs.

PR #406, opened 2022-09-29
refactor: add learn.py

### LINKED ISSUE #392: Add learn.py
Refactor the Draco-learn notebook and extract the relevant logic and move it to the the Draco package itself. This makes it more reusable.

JunranY:
close #392


JunranY:
Added learn.py and data_utils that turns data pairs to training input. 
modified draco_learn notebook accordingly
fixed the math expression in doc

**Problem**: code coverage: I wasn't sure what test cases to write for draco-learn?

draco/data_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        violations = draco.count_preferences(spec)
+        if violations is not None:
+            processed_specs[key] = violations
+    return processed_specs[key]
+
+
+def pairs_to_vec(
+    specs: Dict[str, Dict], fields: Tuple[str, str] = ("negative", "positive")
+) -> pd.DataFrame:
+    """given specs, convert them into feature vectors."""
```
domoritz, line 125: Make this a proper sentence

draco/data_utils.py
```diff
@@ -0,0 +1,127 @@
...
+
+from .draco import Draco
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+draco = Draco()
+
+
+def get_nested_index(fields=None):
```
domoritz, line 18: Can you add types?

draco/learn.py
```diff
@@ -0,0 +1,231 @@
...
+        clf.fit(X_train, y_train)
+
+        if not quiet:
+            print("Train score: ", clf.score(X_train, y_train))
+            if test_size > 0:
+                print("Dev score: ", clf.score(X_dev, np.ones(len(X_dev))))
+
+        return clf
+
+    def train_and_plot(self, test_size: float = 0.3):
```
domoritz, line 60: We should separate the training from the plotting. This way people can use only the training methods which is what most people will want. Moreover, we can then test only the training method. The plotting could either live in a notebook (I think that makes sense given that it's not polished yet) or I a separate method that we mark as not tested.

draco/learn.py
```diff
@@ -0,0 +1,231 @@
...
+from matplotlib.colors import ListedColormap
+from sklearn import svm
+from sklearn.decomposition import PCA
+from sklearn.model_selection import train_test_split
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class DracoLearn:
```
domoritz, line 15: I don't know whether this needs to be a class vs just a collection of methods. Is there benefit to having a class?

domoritz (changes requested):
Remove

draco/data_utils.py
```diff
@@ -0,0 +1,146 @@
...
+        if violations is not None:
+            processed_specs[key] = violations
+    return processed_specs[key]
+
+
+def pairs_to_vec(
+    specs: Dict[str, Dict], fields: Tuple[str, str] = ("negative", "positive")
+) -> pd.DataFrame:
+    """
+    Given pairs of positive and negative draco specification, count the number of times
```
domoritz, line 131:
```suggestion
    Given pairs of positive and negative Draco specifications, count the number of times
```

domoritz (approved):

PR #431, opened 2022-10-30
fix: binder requirements (#430)

### LINKED ISSUE #430: Binder link doesn't have necessary modules
When I launch binder from the readme and then run `debug_draco.ipynb`, I get an error. 

Screen Shot 2022-10-29 at 23 47 11

It looks like the dependencies weren't installed automatically.

peter-gy:
### Goals

- Fix missing Binder modules

### Notes

While Binder supports [various simpler configuration files](https://mybinder.readthedocs.io/en/latest/using/config_files.html), using a `Dockerfile` has the most benefits for us IMHO.

- If we wanted to use a `requirements.txt` we would need to constantly sync between `pyproject.toml` and `requirements.txt` using `poetry export`. While this would be possible using a GH action, it would have been more work than the committed `Dockerfile`
- Using any of the other config options would not fit into our Poetry setup
- With the custom Docker env I declared, opening the terminal in Binder's Jupyter Lab will give access to the same dev environment we are using. That is, executing `make` targets is also possible.

I had to add Draco 2 as a pip package to the Docker environment, otherwise, the notebooks complain about it as a missing dependency.

You can preview this branch in Binder using the following link: [https://mybinder.org/v2/gh/cmudig/draco2/fix/430-binder-requirements](https://mybinder.org/v2/gh/cmudig/draco2/fix/430-binder-requirements)

closes #430

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,11 +18,13 @@ RUN python -m pip install --upgrade pip &&  \
 COPY pyproject.toml poetry.lock ./
 # Purposely not using --no-dev here to install dev dependencies needed for Jupyter Notebooks
 RUN poetry config virtualenvs.create false && \
-    poetry install && \
-    pip install git+https://github.com/cmudig/draco2.git#egg=draco  # Install Draco2 from git, needed for notebooks
+    poetry install
 
 # Copy the project source code
 COPY . .
+
+# Install Draco2 from local sources, needed for notebooks
+RUN pip install -e .
 
 # Grant permissions to the notebook user
 RUN chown -R ${NB_USER} ${HOME}
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN adduser --disabled-password \
     --gecos "Default user" \
     --uid ${NB_UID} \
     ${NB_USER}
-WORKDIR ${HOME}
+WORKDIR ${HOME}/app
 
 # Install dependencies using Poetry
 RUN python -m pip install --upgrade pip &&  \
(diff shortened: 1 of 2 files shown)


peter-gy:
Note that `make test` will fail inside Binder's Jupyter Lab environment with a seg fault. This is most likely caused by the memory limit set for each container running on Binder. However, this should not be an issue at all, as visitors are expected to only explore the interactive docs.

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+WORKDIR ${HOME}/app
+
+# Install dependencies using Poetry
+RUN python -m pip install --upgrade pip &&  \
+    pip install poetry
+COPY pyproject.toml poetry.lock ./
+# Purposely not using --no-dev here to install dev dependencies needed for Jupyter Notebooks
+RUN poetry config virtualenvs.create false && \
+    poetry install && \
+    pip install git+https://github.com/cmudig/draco2.git#egg=draco  # Install Draco2 from git, needed for notebooks
```
domoritz, line 22: Is that better than installing the sources from this repo directly?

domoritz (approved):

Dockerfile
peter-gy: The only reason I chose this approach instead of installing locally from source using `python -m pip install -e .` is that I wanted to have all the dependency installation commands defined in one layer, before copying the source code with `COPY . .`. But as I think about it, we would spare the `git clone` networking with the local installation and we would ensure that package contents we access with `import draco` are actually sourced from the latest version of the code, instead of a maybe-cached dependency layer.

peter-gy: I addressed this in a6e9ff4, thanks for the catch!

PR #438, opened 2022-11-01
feat: default Vega-Lite-based renderer (#342)

### LINKED ISSUE #342: Create a renderer for visualization specifications
Draco isn't just targeting a specific language but it could still make sense to have a default renderer based on Vega-Lite.

peter-gy:
### Goals

- Create a default renderer based on Vega-Lite
  - Used Altair to get automatic VL schema validation
  - The renderer returns an `alt.VConcatChart | alt.HConcatChart | alt.FacetChart | alt.Chart`, ready to be displayed, exported, customized, etc.
- Define the schema of the Draco specification language using Pydantic
  - Using raw dicts one has to fight a lot with the type checkers
  - Having typed access to the schema makes its traversal and processing more readable

closes #342
spawns #449
spawns #455


peter-gy:
### Vega-Lite vs. Draco spec

Below I list my takeaways in a nutshell from implementing all the example charts from the docs ([https://dig.cmu.edu/draco2/facts/examples.html](https://dig.cmu.edu/draco2/facts/examples.html))  in Vega-Lite using Altair.

- VL **mostly** covers all our rendering needs, in most cases the mapping from our schema to the VL schema is straightforward ✓
- VL has no explicit support for toggling between cartesian and polar coordinate systems
  - However, we can use an `arc` mark & the encodings `theta`, `radius` to overcome this ✓
- As far as I can tell, VL does not allow for nested faceting / repeating / concatenation. This is problematic since Draco can produce multi-view recommendations and each view might or might not be faceted 🚫

### Questions regarding Draco's spec

- What do we consider as a strictly valid spec? Defining this would be important so that the renderer method can validate the input before trying to convert it into VL. For example, the following spec ([Pie Chart Example](https://dig.cmu.edu/draco2/facts/examples.html#pie-chart)) does not define a `field` for its encoding on the `y` channel:

```python
{'field': [{'name': 'temperature', 'type': 'number'},
           {'name': 'condition', 'type': 'string'}],
 'number_rows': 100,
 'view': [{'coordinates': 'polar',
           'mark': [{'encoding': [{'aggregate': 'count',
                                   'channel': 'y',
                                   'stack': 'zero'},
                                  {'channel': 'color', 'field': 'condition'}],
                     'type': 'bar'}],
           'scale': [{'channel': 'y', 'type': 'linear', 'zero': 'true'},
                     {'channel': 'color', 'type': 'categorical'}]}]}
```

While it is easy to infer that the `temperature` field should be used here, **is the renderer expected to handle this inferring**?

- What do you think, would it make sense to define a Draco spec strictly? We could use Pydantic for this & the implementation would mean relatively little effort. It has an awesome DX when it comes to defining schemas (see [https://pydantic-docs.helpmanual.io/usage/schema/](https://pydantic-docs.helpmanual.io/usage/schema/)) and we already depend on the package.

@JunranY, @Zehua-Zeng, @domoritz

- do you know a workaround to concatenate faceted views in VL?
- what do you think about introducing a Pydantic-based schema validation?
- in general, what kind of development roadmap would you prefer for this feature? Should we...
  - merge an MVP as soon as possible that is capable of rendering simpler recs (basically any non-multi-view rec in a cartesian polar system) and work on further refinements iteratively, or
  - create a more elaborate issue from this handling all specs & create a rendering base-class to support multiple renderer implementations?

peter-gy:
Hey @domoritz! While the code I pushed is indeed not really LGTM (as it is partial work full of quick & dirty experimenting), I believe the [LGTM.com](https://lgtm.com/projects/g/cmudig/draco2/rev/pr-4548ce305475502e2ba9f72addf6a6d009fec08e) report is not right:

Screenshot 2022-11-08 at 17 37 17

`kw_only` is a totally valid parameter name since Python `3.10`. Is it possible that the Python version based on which LGTM.com checks our code is not in sync with the version we actually use? I did not update it in 06144c472ad2daf036263f64f0ec158a30711416, as I have no access to the job's config as far as I know.

peter-gy:
### Update

Writing tests for the renderer are still in progress. All charts defined with a `cartesian` coordinate system are supported by the renderer. It produces a syntactically valid VL spec for each of the example charts. It is possible that the VL output still needs to be fined-tuned semantically.

A review round on the implemented approach would be much appreciated to find out in what other ways should the renderer be further-developed.

draco/renderer/altair_renderer.py
```diff
@@ -0,0 +1,263 @@
...
+                match channel:
+                    case "row":
+                        chart = chart.facet(row=alt.Row(**facet_args))
+                    case "col":
+                        chart = chart.facet(column=alt.Column(**facet_args))
+                    case _:
+                        raise ValueError(f"Unknown facet channel: {channel}")
+        return chart
+
+    def __visit_mark(self, ctx: MarkContext) -> VegaLiteChart:
```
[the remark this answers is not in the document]
peter-gy: @domoritz I would dismiss this CodeQL alert. Adding a `return None` to the end of the method would be just the addition of an unreachable line. What do you think?

peter-gy:
## Still Open

- Handling decision between `vconcat` and `hconcat` (`vconcat` is used by default at the [moment](https://github.com/cmudig/draco2/pull/438/commits/5c534861fc223229bfb0e80ef8451632fb9c6122))
- Handling charts in polar coordinate-system

peter-gy:
@domoritz could you please take a look at this PR?

draco/types.py
```diff
@@ -1,3 +1,315 @@
...
+from typing import Iterable, Literal, TypeAlias
 
+# granular imports to satisfy `pyright`
+import pydantic.class_validators as pydantic_validators
+import pydantic.config as pydantic_config
+import pydantic.fields as pydantic_fields
+import pydantic.main as pydantic_main
+import pydantic.types as pydantic_types
+
+"""
```
domoritz, line 10: As discussed, this file is Vega specific and should be moved into the Vega renderer.

domoritz (changes requested):

peter-gy:
The failing workflows might be related to actions/setup-python#162.

draco/renderer/altair_renderer.py
```diff
@@ -0,0 +1,434 @@
...
+        :param ctx: The current visitor context.
+        :return: The chart with the mark applied.
+        :raises ValueError: if the mark type is not supported
+        """
+        chart, encodings, mark_type = (ctx.chart, ctx.mark.encoding, ctx.mark.type)
+        match mark_type:
+            case "bar":
+                encodes_x_and_y = all([e.channel in ["x", "y"] for e in encodings])
+                if encodes_x_and_y:
+                    return chart.mark_arc(stroke="#ffffff") + chart.mark_text(
```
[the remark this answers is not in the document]
peter-gy: The mental notes I took when writing this piece of code was the following: 

_When rendering a bar chart in a polar coordinate system, we are essentially rendering a classical radial chart in a cartesian coordinate system, using `theta` and `radius` as the channels instead of the polar `x` and `y` channels._

TLDR: I added a stroke here so there is a visual separation between the slices, just as we would have gaps between bars in a bar chart.

**Without Stroke**



**With Stroke**

draco/renderer/altair_renderer.py
```diff
@@ -0,0 +1,434 @@
...
+                    return chart.encode(theta=alt.Theta(**encoding_args))
+            case "color":
+                return chart.encode(color=alt.Color(**encoding_args))
+                # Should never happen, a pydantic error would be raised sooner
+            case _:  # pragma: no cover
+                raise ValueError(
+                    f"Unknown channel: {encoding.channel}"
+                )  # pragma: no cover
+
+    def __get_field_type(self, fields: list[Field], field_name: FieldName) -> str:
```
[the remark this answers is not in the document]
peter-gy: Meeting notes:

Scale type provides a cleaner mapping, as it makes it possible to differentiate between different number types such as numerical, ordinal, and categorical.

draco/renderer/altair/altair_renderer.py
```diff
@@ -392,26 +393,125 @@ def __visit_encoding_polar(self, ctx: EncodingContext) -> VegaLiteChart:
...
+        # Multi-criteria lookup to determine the type of the field
+        # based on the scale type (if any) AND the raw data type.
         renames = {
-            "number": "quantitative",
-            "string": "nominal",
-            "boolean": "nominal",
-            "datetime": "temporal",
+            "linear": {
+                "number": "quantitative",
+                "string": "nominal",
```
domoritz, line 471: string and boolean are not supported for linear, no?

draco/renderer/altair/altair_renderer.py
```diff
@@ -392,26 +393,125 @@ def __visit_encoding_polar(self, ctx: EncodingContext) -> VegaLiteChart:
...
+                "boolean": "nominal",
+                "datetime": "temporal",
+            },
+            "log": {
+                "number": "quantitative",
+                "string": "nominal",
```
domoritz, line 477: same as above

domoritz (approved):
Just one last comment.

peter-gy:
@domoritz, I added the discussed changes in bd7ca3b

draco/renderer/altair/altair_renderer.py
```diff
@@ -0,0 +1,532 @@
...
+                "number": "quantitative",
+                "datetime": "temporal",
+            },
+            "log": {
+                "number": "quantitative",
+                "datetime": "quantitative",
+            },
+            "ordinal": {
+                "number": "ordinal",
+                # Accounting for colors
```
domoritz, line 479: What does this mean? I think we can still use ordinal for now.

domoritz (approved):

draco/renderer/altair/altair_renderer.py
peter-gy: Thanks for pointing this out. As soon as I adjusted this to `ordinal` I discovered a bug after seeing that the unit tests are failing 🙃

It is fixed now.

PR #464, opened 2023-01-10
docs: how to use learned weights (#418)

### LINKED ISSUE #418: Document how to use learned weights
Show how the weights can be used in Draco to apply the model for recommendation.

JunranY:
closes #418


docs/applications/draco_learn.ipynb
```diff
@@ -713,9 +634,128 @@
...
+      "#show attribute/3.\n",
+      "\n",
+      "OUTPUT:\n",
+      "{'field': [{'name': 'temperature', 'type': 'number', 'unique': 100}],\n",
+      " 'number_rows': 100,\n",
+      " 'task': 'summary',\n",
+      " 'view': [{'coordinates': 'polar',\n",
+      "           'mark': [{'encoding': [{'binning': 10,\n",
+      "                                   'channel': 'x',\n",
+      "                                   'field': ('f', 0)},\n",
```
peter-gy, line 717: I would not expect to see `'field': ('f', 0)` here, `'field': 'temperature'` would make more sense in this context. Is  [histogram.lp](https://github.com/cmudig/draco2/blob/635085365cd5d0e23b8267424b0dec5ab998eead/draco/asp/examples/histogram.lp) in the correct format?

peter-gy (commented):
@JunranY  The docs are clear and easy to follow. Displaying a rendering of the `hist_spec` would be nice, but the output specification seems to be invalid to me. Could you please adjust it?

docs/applications/draco_learn.ipynb
JunranY: As I remember, we had to use the `(x,0)` format in the generator to generate certain number of encodings etc. I can't remember if that's required for fields, though. It's refer to the entity id as `temperature` is for the name.

JunranY: Oh, do you actually mean the output dict format?

peter-gy: Exactly, I was referring to the final output dict. In its current form it cannot be rendered. If this spec is valid, then I should adjust the renderer to handle it, otherwise we should explore why an invalid spec is produced.

JunranY: Thanks! We will address this in a separate pr for issue #467

peter-gy: Perfect, thanks a lot!

PR #466, opened 2023-01-11
chore: switch to ruff

domoritz:
Depends on https://github.com/charliermarsh/ruff/issues/282


peter-gy:
@domoritz, we discussed at one of our meetings that we are running lint checks both as part of the pre-commit check and the `test.yml#lint` job. Can we get rid of the latter?

domoritz:
Yes, we can. Didn't we already remove it?

domoritz:
Replaced by https://github.com/cmudig/draco2/pull/507

PR #468, opened 2023-01-13
fix: use field name as reference in encodings (#467)

### LINKED ISSUE #467: Use field name as reference in encodings
Instead of requiring to use the id (which may be hidden in a spec), use the name of the field. Also add a constraint that fields should be unique?

JunranY:
closes #467 

write helpers to match `field` entities by name in the encoding. 
todos are:
- [x] all `attribute((encoding,field),,)` in soft, hard and tests
- [x] in the helpers
- [x] `constraint.lp`: fields need to be defined ahead
- [x] `define.lp`: valid domain for field names
- [x] example specs
- [x] check for facet fields, and repeat previous steps
- [x] hard constraint for unique field names
- [x] documentations


domoritz:
Can wer simplify https://dig.cmu.edu/draco2/facts/examples.html now and remove the `__id__`?

PR #480, opened 2023-01-25
chore: update Python version in `Dockerfile` & document dev with docker

peter-gy:
As we are emphasizing the tooling and the DX around Draco 2, I thought it would be useful to have a copy-paste-ready command that anyone can execute with Docker, to start all the components locally.


peter-gy:
Spawned #483

PR #482, opened 2023-01-27
fix: address scale translation warnings (#472)

### LINKED ISSUE #472: Fix scale translation warnings in the renderer

peter-gy:
Adjusts the renderer to let Vega-Lite infer the scale type of a chart whenever possible.

Closes #472


draco/tests/renderer/altair/test_altair_renderer.py
```diff
@@ -875,7 +875,7 @@ def test_facets(spec: dict, expected_vl: dict, renderer: AltairRenderer):
                 "x": {
                     "field": "condition",
                     "type": "ordinal",
-                    "scale": {"type": "ordinal"},
+                    "scale": {},
```
domoritz, line 878: Do we need to generate an empty object here? Would be nice to have cleaner outputs.

domoritz (approved):

draco/tests/renderer/altair/test_altair_renderer.py
peter-gy: Fixed in 09de654

PR #484, opened 2023-01-27
ci: test `Dockerfile` (#483)

### LINKED ISSUE #483: ci: test `Dockerfile` in GitHub workflows
We should test that the docker environment works as expected by running `make all` inside a GH workflow.

peter-gy:
The `test-docker` job I added in this PR executes `make all` inside a docker container based on our root `Dockerfile`. If the `test-docker` job succeeds along with the other, non-dockerized checks, we can conclude that the dockerized environment works as expected.

The job is expected to fail if any of the other jobs (`lint` or `test`) fail.

I had to update from debian buster to bullseye, since I kept getting a Python seg fault inside the container running on the GH runner. The error disappeared after upgrading to bullseye.

Closes #483

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,8 +44,8 @@ jobs:
       - name: Build docker image
         run: docker build -t draco .
 
-      - name: Run `make all`
-        run: docker run --rm draco bash -c "make all"
+      - name: Run all make targets
+        run: docker run --rm draco bash -c "make"
 
   lint:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -36,6 +36,17 @@ jobs:
...
+  test-docker:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build docker image
+        run: docker build -t draco .
+
+      - name: Run `make all`
+        run: docker run --rm draco bash -c "make all"
```
domoritz, line 48: remind me, do we need to say "all"?

```suggestion
        run: docker run --rm draco bash -c "make"
```

domoritz (approved):

.github/workflows/test.yml
peter-gy: No, we don't, thanks for the hint!

PR #507, opened 2023-02-22
chore: switch to ruff for linting

domoritz:


peter-gy (approved):
I am pretty excited about switching to ruff, it's impressive how fast it is.

PR #510, opened 2023-02-28
docs: add more to the debugging guide

JunranY:
Meanwhile, trying to reduce the number of redundant charts from generator. 
It's not fully resolved for charts with symmetry Draco spec.


domoritz (approved):

JunranY:
@domoritz further question for breaking Draco spec structure symmetry:

I was trying to add the constraints to make generated channels following the certain order, but there are a few questions I couldn’t wrap my head around. For example, if the input defines one encoding e0 and the generator adds more, then the channel of e0 should follow the order if its field is not defined, otherwise it should be able to be anything (see example down below). The questions is that to write the constraint, how do we check what’s included in the input and what’s from the generator. Same problem may also happen with mark types: they should follow the order if the input contains only the ids as placeholder, but not if their encodings are also defined in the input. 

All of the above makes me think if we should check for redundancy not by using constraints but checking the dictionary format after clingo spits out the answers. 

```prolog
attribute((encoding,field),(e,0),temp_max). 
attribute((encoding,channel),(e,0),x).
attribute((encoding,field),(e,1),precipitation).
attribute((encoding,channel),(e,1),y).
```
and 
```prolog
attribute((encoding,field),(e,0),temp_max). 
attribute((encoding,channel),(e,0),y).
attribute((encoding,field),(e,1),precipitation).
attribute((encoding,channel),(e,1),x).
```
are not the same, and the latter should be allowed when the fields are defined in the input.

Separate from this, I was trying a constraint that partially reduce the symmetry if input ids are defined differently from in the generator:
```prolog
violation(enforce_order):- 
    entity(view,root,V), M1 < M2, 
    attribute((mark,type),(V,M1),T1), 
    attribute((mark,type),(V,M2),T2), 
    not T1 < T2. 
```
but it seems to increase the grounding size by a little bit.

domoritz:
Good question. I don't think there is a way to check whether a fact is coming from the generator or the user program but maybe there is. It would be better to break symmetries in clingo because it will speed up constraint solving significantly.

domoritz:
Do you want to add a test for the grounding size or leave this as is for now?

domoritz (approved):

JunranY:
I'm trying to add the tests for grounding size, but it seems that clingo api doesn't support the "--text" argument. 
I found a way to [execute modules](https://docs.python.org/3/library/runpy.html#runpy.run_module), but cannot figure out how to pass in the lp files.

domoritz:
Ah okay. We can look at that later then. Maybe we need to call the cli. Anyway, we can merge this pull request as is.

PR #562, opened 2023-04-06
chore: update python version, pyodide and jupyterlite

peter-gy:
### Goals

- Use jupyter-lite stable `0.1.0`
- Use Pyodide `0.23.0`
- Use Python `3.10.11`

### Notes

We can't update to Python `3.11.*`, because of google/pytype#1308


peter-gy:
Python `3.10.11` was released on the 5th of April, that is, yesterday ([source](https://www.python.org/downloads/release/python-31011/)). `actions/setup-python@v4` does not support this latest version just yet, therefore we have failing CI jobs.

PR #602, opened 2023-05-15
chore: migrate to altair 5.0.0

peter-gy:
Extends #599 with test-case-fixing and Pyodide-altair version bump.


domoritz:
Thank you.

PR #667, opened 2023-07-30
chore: add debug chart title

peter-gy:
Uses `Constraint` as the x-axis label of the debug chart. Addresses a suggestion of the primary reviewer.

visualization


peter-gy:
Also fixes the issue that the weights get summed up:

Screenshot 2023-07-30 at 19 34 43

PR #680, opened 2023-08-05
docs: finalize docs before releasing `2.0.0`

peter-gy:


pyodide/README.md
```diff
@@ -24,7 +24,7 @@ The version of the release needs to match exactly the version of this package.
 You can automate this using the following commands:
 
 ```shell
-DRACO_PYODIDE_VERSION=2.0.0b4  # Replace with the package version of your choice
```
domoritz, line 27: I think we need to exclude this change before the release.

domoritz: Oh wait, that's just in the docs. That's fine then.

PR #748, opened 2023-10-26
chore: migrate to `ruff format` from `black`

peter-gy:


Makefile
```diff
@@ -13,7 +13,7 @@ cover:
 .PHONY: lint
 lint:
 	@echo "==> 👕 Linting"
-	@poetry run black .
+	@poetry run ruff format draco jupyterlite pyodide docs
 	@poetry run ruff .
```
domoritz, line 17: Can we delete this line now?

domoritz (approved):

Makefile
peter-gy: You mean `poetry run ruff .`? I think that's still necessary to do the actual linting.

domoritz: I see. Because formatting is only catching a subset of things. Sounds good.

Issue #791, opened 2024-01-24
Bad recommendation using log scale with zero


domoritz:
In https://dig.cmu.edu/draco2/applications/debug_draco.html#how-to-modify-the-knowledge-base-with-constraints, we recommend log scale with zero or binning that doesn't really make sense. 

- [ ] Make sure we never recommend log with zero
- [ ] Don't recommend log when extent includes zero

<img width="1132" alt="Screenshot 2024-01-24 at 11 06 09" src="https://github.com/cmudig/draco2/assets/589034/b11d9f94-298c-415e-a4ec-925605553d38">

domoritz:
Done in https://github.com/cmudig/draco2/pull/792

PR #792, opened 2024-01-30
feat: add hard constraints on log scale and zero scale

JunranY:
for #791.
there is already a hard constraint for not using log with zero or negative values, here are the additional things I did: 
- added a hard constraint to make sure not to recommend log scale when extent includes zero
- To avoid redundant recommendation, I added a hard constraint to not recommend zero scale when extent includes zero
- added the facts about data schema to the query because previously the min and max are not both included
- updated the "revising soft constraints" example to cooperate the above change


domoritz (approved):
Thank you! 

> To avoid redundant recommendation, I added a hard constraint to not recommend zero scale when extent includes zero

I don't get this. I think it should be possible to specify `zero: true` on a scale even if the data already includes zero. This is actually a common case, I would think.

JunranY:
I thought the definition of `zero: true` is "scale starts with 0" (as in the description), which means if the extent spans across 0 there is no way the scale is starting with 0.

domoritz:
Hmm, I was going off of Vega which says "Boolean flag indicating if the scale domain should include zero.". We could rewrite the comment here or leave the current comment and accept your changes (probably the easiest option). 

Alternatively, we can just remove the flag and make it so that Draco can extend the domain for a scale to include zero but that's a bit more work.

domoritz:
we decided
* zero should be an indicator to the scale domain computation to include zero (in addition to unioning the domains from all the fields that use the scale). 
* zero should have some soft-constraint cost
* not including zero for bar or area charts should have a higher cost

JunranY:
I just realized that all of these are already implemented in: 
- `hard(bar_area_without_zero)`: Bar and area mark requires scale of continuous to start at zero. 
- `soft(continuous_not_zero)`: Prefer to include zero for continuous (binned doesn't need zero). 
- `soft(cross_zero)`: Prefer not to include zero as baseline when the range of data crosses zero. 

I'm thinking about 1) adding a constraint to allow zero only for quantitative scale (i.e. linear), 2)adjust the example in doc accordingly

domoritz:
Sounds good. Let's be careful with terminology since a log scale could be considered quantitative as well.

domoritz (approved):

PR #838, opened 2024-04-17
fix: false positive pyodide build cache hits

peter-gy:
Previously we used the built Pyodide package version as the cache key exclusively.

This yielded false positive cache hits when only the source code of a module changed (the sha256 signature), but not its version, as no official release took place. Therefore, we kept reusing built modules even though they were obsolete.

This PR introduces a fix by using `{version}@{sha256}` as the hash key.
Technically, `{sha256}` would be sufficient as a key, but including the version aids debugging.


domoritz:
Is this still relevant?

peter-gy:
It is. I will need to dramatically simplify the pyodide build flow

peter-gy:
Working on this as part of #899.

PR #883, opened 2024-08-20
chore: update lockfile

domoritz:


peter-gy:
Will be addressed once #899 is merged.

domoritz:
Okay, closing.

PR #899, opened 2024-11-10
refactor: migrate repository to `uv` & simplify Pyodide build

### LINKED ISSUE #884: Using AltairRenderer in pyodide enviornment
Hello, I'm currently working on a project using draco-pyodide.

I imported AltairRenderer to convert the recommended visualization to Vega-Lite, but I'm encountering the following error. (This error appears not only in the pyodide console you provided but also in my source code.)

image

In the released tarball, the version of pydantic is 2.4.2, but as shown in the attached image, version 1.10.7 of pydantic is loaded in the environment. I think this might be the cause of the problem.

How can I resolve this issue?

Thank you for developing such an amazing library.

### LINKED ISSUE #895: Consider switching to UV
Consider using https://github.com/astral-sh/uv rather than poetry.

peter-gy:
## How to test this PR?

`git checkout` this branch, then execute `make`. Everything should work without any further action (linting, tests, typechecks, doc-gen).

Running all the `make` targets still takes a while, because `mypy` and `pytype` are really slow and generating the Jupyter Book output also requires patience, given that we execute all notebooks on the fly to generate the HTML bundle.

> [!NOTE]  
> Please make sure you have `uv` installed and available on your `PATH`.
> You can refer to the installation instructions here: https://docs.astral.sh/uv/getting-started/installation/

## CHANGELOG

- Drop Poetry in favor of uv
- Move `draco` module source under `src/`
- Use Python 3.12.7
- Update `.github/workflows/test.yml` CI flow
- Make devcontainer `Dockerfile` slimmer (~2x image build performance improvement)
- Define `[project.optional-dependencies]` in `pyproject.toml`, allowing `pip install draco` to pull only core dependencies
  - `pip install 'draco[server]'`: will include FastAPI deps
  - `pip install 'draco[renderer]'`: will include deps to render chart specs
  - `pip install 'draco[learn]'`: will include `scikit-learn` enabling `draco.learn.train_model`
- Stop maintaining custom Pyodide distribution, since `clingo` and `vega-datasets` are now members of the official Pyodide distribution (available through https://cdn.jsdelivr.net/pyodide/<verrsion>/full/)
- Build JupyterLite into docs via [jupyterlite-sphinx](https://pypi.org/project/jupyterlite-sphinx/).

## Pyodide Build Simplifications

The complexity of the current Pyodide build pipeline stems from the fact that `clingo` is an impure Python package (depends on `cffi`) without an official recipe for it in the Pyodide repo. In parallel to this PR I am opening PRs against https://github.com/pyodide/pyodide to include packages we need into their official build. This way, we can dramatically reduce the code we maintain in this repo responsible for WASM builds.

### Foreign PRs

- https://github.com/pyodide/pyodide/pull/5183 
- https://github.com/pyodide/pyodide/pull/5184

## Closes

Closes #895
Closes #884


src/draco/tests/test_data_utils.py
```diff
@@ -10,7 +10,7 @@
 draco: Draco = Draco()
 learn_data: dict[str, dict[str, str]] = {}
 
-root_path: Path = Path(__file__).resolve().parents[2]
+root_path: Path = Path(__file__).resolve().parents[3]
```
peter-gy, line 13: We moved the module sources under `src/` so we must look one parent ahead.

src/draco/tests/test_learn.py
```diff
@@ -4,7 +4,7 @@
 from draco.data_utils import pairs_to_vec
 from draco.learn import train_model
 
-root_path = Path(__file__).resolve().parents[2]
+root_path = Path(__file__).resolve().parents[3]
```
peter-gy, line 7: We moved the module sources under `src/` so we must look one parent ahead.

src/draco/tests/test_data_utils.py
domoritz: What's the point of moving the source? There is only one folder so it seems unnecessary.

peter-gy: We will still need some minimal tooling around Pyodide and JupyterLite. I planned to move those files under `src/`.
Do you suggest to keep it top-level.

domoritz: Let me see the end result.

docs/jupyter-lite.json
```diff
@@ -5,7 +5,7 @@
     "appUrl": "./lab",
     "litePluginSettings": {
       "@jupyterlite/pyodide-kernel-extension:kernel": {
-        "pyodideUrl": "/static/pyodide/pyodide.js"
+        "pyodideUrl": "https://cdn.jsdelivr.net/pyodide/dev/full/pyodide.js"
```
peter-gy, line 8: For now we must use the `dev` version (rolling release pointing to `pyodide@main`) because it is the only distro over a public CDN which already includes https://github.com/pyodide/pyodide/pull/5183  and https://github.com/pyodide/pyodide/pull/5184

src/draco/tests/test_data_utils.py
peter-gy: I ended up dropping `src/`, as I compressed the JupyterLite-related tooling into a single file (`docs/jupyterlite_nb_patch.py`)

docs/jupyter-lite.json
domoritz: Sounds good

domoritz (approved):
Thanks

